PR #5, opened 2016-03-30
Replace all the instances instead of the first one

obastemur:
Current implementation fails to replace additional instances of esc. characters.


Stevenic:
nice catch :)

Stevenic:
I'll need to apply the same fix to the SkypeBot as well. Thanks

obastemur:
Just saw it. I could wrap them into a single commit. Can fix that part too ?

Issue #13, opened 2016-03-31
Bot Builder for NodeJs - error in example


Aleksion:
Hi,

I apologize if I'm reporting in the wrong place - but this had be stumble when I first tried out the code.

The example on: http://docs.botframework.com/builder/node/guides/core-concepts/#publishing-to-the-bot-connector-service

has an error in the code: it uses a textBot even though it should be using a BotConnectorBot.

The text states this correctly, however, the coding example doesn't

Stevenic:
Thanks for the fix... I merged it :)

PR #14, opened 2016-03-31
Updated publish code example

Aleksion:
Updated the publish code example from TextBot to BotConnectorBot as metnioned in #13

### DIFF
--- a/Builder/Node/guides/core-concepts.md
+++ b/Builder/Node/guides/core-concepts.md
@@ -147,7 +147,7 @@ var builder = require('botbuilder');
 
 var server = restify.createServer();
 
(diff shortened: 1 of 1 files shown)


Stevenic:
Nice catch

Builder/Node/guides/core-concepts.md
```diff
@@ -147,7 +147,7 @@ var builder = require('botbuilder');
 
 var server = restify.createServer();
 
-var helloBot = new builder.TextBot();
+var helloBot = new builder.BotConnectorBot();
```
mike2real: Hey

PR #40, opened 2016-04-02
Fix for EntityRecognizer parseBoolean

beebes:
- yesExp and noExp regexes used incorrect "\z" character to denote end of line anchor, in JavaScript they should use "$"
- consequence of this was that Prompts.confirm would always report undefined
- added EntityRecognizer test with parseBoolean tests


Stevenic:
Thanks... I actually just dropped the end of line designator. I decided to make it more flexible and leave it open for things like "sure that's fine"

beebes:
awesome, thanks!

PR #41, opened 2016-04-03
CHORE - Remove Trailing Spaces

clickthisnick:
Removes trailing spaces
Saves 0.008458 megabytes


Stevenic:
These are auto generated files so the spaces will just come back.

PR #71, opened 2016-04-07
PromptConfirm TryParse should ignore casing on message

pagebrooks:
When replying to a PromptConfirm, the logic does not properly interpret answers that are not lower-cased.  It is common for smart phones to "autocorrect" response casing so the bug is quite noticeable when interacting with a bot through a smartphone.

### DIFF
--- a/CSharp/Library/Dialogs/PromptDialog.cs
+++ b/CSharp/Library/Dialogs/PromptDialog.cs
@@ -159,7 +159,7 @@ namespace Microsoft.Bot.Builder.Dialogs
 
             protected override bool TryParse(Message message, out bool result)
             {
-                switch (message.Text)
+                switch (message.Text.Trim().ToLower())
                 {
                     case "y":
                     case "yes":


CSharp/Library/Dialogs/PromptDialog.cs
```diff
@@ -159,7 +159,7 @@ public PromptConfirm(string prompt, string retry, int attempts)
 
             protected override bool TryParse(Message message, out bool result)
             {
-                switch (message.Text)
+                switch (message.Text.Trim().ToLower())
```
sluu99: Should we use ToLowerInvariant()?

msft-shahins: I think it is safer to use ToLowerInvariant() but since we are not localized, it doesn't make a difference.

pagebrooks:
@sluu99 @msft-shahins Thanks! I agree, but I saw many other uses of ToLower() elsewhere in the project (and no uses of ToLowerInvariant()) so I just followed the pattern. Maybe switching over to ToLowerInvariant() should be separate issue?

PR #76, opened 2016-04-08
Fix url encode of text when language is french

swoog:
I have make tests with french language. Accents are correct in LUIS when use UrlEncode().

Example : q=dis moi si je possède lanfeust de troy ?

### DIFF
--- a/CSharp/Library/Luis/LuisService.cs
+++ b/CSharp/Library/Luis/LuisService.cs
@@ -86,13 +86,12 @@ namespace Microsoft.Bot.Builder.Luis
 
         Uri ILuisService.BuildUri(string text)
         {
-            var queryString = HttpUtility.ParseQueryString(string.Empty);
-            queryString["id"] = this.model.ModelID;
-            queryString["subscription-key"] = this.model.SubscriptionKey;
-            queryString["q"] = text;
+            var modelId = HttpUtility.UrlEncode(this.model.ModelID);
+            var subscriptionKey = HttpUtility.UrlEncode(this.model.SubscriptionKey);
+            var q = HttpUtility.UrlEncode(text);
 
             var builder = new UriBuilder(UriBase);
-            builder.Query = queryString.ToString();
+            builder.Query = $"id={modelId}&subscription-key={subscriptionKey}&q={q}";
             return builder.Uri;
         }
 


msft-shahins:
I have tried your query string with current implementation and the http call was successful. Here is builder.Query that I see for the query you mentioned: 

> ?id=c413b2ef-382c-45bd-8ff0-f76d60e2a821&subscription-key=6d0966209c6e4f6b835ce34492f3e6d9&q=dis+moi+si+je+poss%c3%a8de+lanfeust+de+troy

 If you look at the implementation of [HttpUtility.ParseQueryString](http://referencesource.microsoft.com/#System.Web/httpserverutility.cs,a711aeeae301c09c) it returns [HttpValueCollection](http://referencesource.microsoft.com/#System.Web/HttpValueCollection.cs) that its [ToString()](http://referencesource.microsoft.com/#System.Web/HttpValueCollection.cs,ad077f34023fd2eb) implementation do the urlencode on both keys and values. Can you let me know what is the problem that you see with current implementation that you are trying to fix with your pull request?

swoog:
LUIS API have problem to decode "%c3%a8". I have make a test :

            var q = "Possède";

            var queryString = HttpUtility.ParseQueryString(string.Empty);
            queryString["q"] = q;
            Console.WriteLine(queryString.ToString());

            var val1 = HttpUtility.UrlEncodeUnicode(q);
            Console.WriteLine($"q={val1}");

            var val2 = HttpUtility.UrlEncode(q, Encoding.UTF8);
            Console.WriteLine($"q={val2}");

The result is :
q=Poss%u00e8de
q=Poss%u00e8de
q=Poss%c3%a8de

In implementation, I have see the settings aspnet:DontUsePercentUUrlEncoding to fix my propblem. When I use true with this parameter. HttpValueCollection use UrlEncode with UTF8 instead of UrlEncodeUnicode.

Documentation says "defaults to true when targeting >= 4.5.2, otherwise false" but it is always false for me :|

msft-shahins:
Are you sure that your host process (asp.net app) is targeting .NET >= 4.5.2? All of our samples target .NET 4.6 and don't have this issue. Bot.Builder.dll is currently targeting 4.5 and we are thinking about upgrading the target framework for it to prevent this issue.

swoog:
I have search and my project is in .Net 4.5.2 Framework but in web.config I have 
<compilation targetFramework="4.5.2" />
<httpRuntime targetFramework="4.5" />

When I change httpRuntime to 4.5.2 it seems works with UTF8 encode. Thanks

PR #77, opened 2016-04-08
Add score property on entity recommendation

swoog:


msft-shahins:
Thanks for your pull request. This is an auto generated file using azure [autorest](https://github.com/Azure/autorest). We regenerated the file in this [commit](https://github.com/Microsoft/BotBuilder/commit/c026fc71cffa5de4dd88287f3db88c90a17bb5e5) to include all the missing fields. Our next release (merge to master) will include these set of changes.

PR #134, opened 2016-04-18
Make Form preset entity validation async

elliatab:
Reproduction steps:
- Add a ValidateDelegate performing an asynchronous task to a Form Field Step (an http call in my case)
- Pre-fill the Step with a value

The form dialog will fail to start.


chrimc62:
Thanks!  This was causing a deadlock so it is a good fix.

PR #136, opened 2016-04-18
Changed Math.round to Math.floor to fetch random array element

FedericoElles:
Nice feature introduced in 0.9.0. Just tested this and got an "undefined" as bot reply. Found this cause.


Stevenic:
Good catch.

PR #148, opened 2016-04-19
message.setText("string") did not work

FedericoElles:
The parameters "msg" and "prompts" might be a string, so this is considered now. Before, prompts seemed to be undefined even while msg was defined, so m.text was set to undefined. Not sure if this solution is the best.


Stevenic:
So there is a bug here but the fix is different. There's not actually a "msg" variable passed in, just prompts. This is typescript so when you have multiple function signatures like that only the last one is actually the real signature. The others are used just for type checking.

Thanks for finding the bug. I'll fix it.

Issue #175, opened 2016-04-23
Dialog localization


marcoamendola:
I noticed that the PromptDialog.Confirm dialog only understands replies in English. I would like to contribute the Italian localization. Are you accepting pull requests in this area?

wiltodelta:
And how about add yep, nope to Dialog in English?

chrimc62:
The most recent release adds a resource file for all strings.  We would love to get contributions to localize that resource file.  I am also working on making it so that you can generate a resource file from the automatically generated stuff in FormFlow.  (I actually have a good chunk of the code checked-in, but it is disabled for now.)  I will also add yep and nope to the resource file, but it is also easy to override the matched terms by changing FormConfiguation to reflect what you think should be there for your program.

PR #189, opened 2016-04-27
IT localization

marcoamendola:
Contains Italian translation of the resource file used to build dialogs' phrases (see issue #175).
The style of the Italian translation is slightly more "formal" than the English version, in order to address grammatical gender and number without introducing an inflection engine.


chrimc62:
I did not merge it directly because I'm revamping how we handle the localization work flow, but I did pull your file into the new system.  It will be in the next release.  Thanks for doing the work!

PR #198, opened 2016-04-28
Links to the samples in the readme file

sebagomez:
is just something I found useful while reding the readme file a direct way of getting to the actual sample


chrimc62:
Thanks!

PR #282, opened 2016-05-13
Remote emulator endpoint

mikkhait:
I'm working on a Mac and emulator is running on a VM. I want to be able to specify a remote endpoint for an emulator as well.

(sorry about line endings in ts)


Stevenic:
Nice... Thanks for fixing :)

Stevenic:
Actually... I just noticed an issue. There's always an endpoint. I'll need to modify this.

PR #284, opened 2016-05-14
Changes in the LuisService initialization

evnik:
* Storing of LUIS service URI query parameters in a prebuilt `HttpValueCollection` object rather than building that object from stratch on each query to LUIS.
* A way to create an instance of `LuisService` without `LuisModelAttribute`, that allows to keep LUIS parameters in AppSettings (or some other place), rather than hardcode them.

### DIFF
--- a/CSharp/Library/Luis/LuisService.cs
+++ b/CSharp/Library/Luis/LuisService.cs
@@ -75,10 +75,20 @@ namespace Microsoft.Bot.Builder.Luis
         /// </summary>
         /// <param name="model">The LUIS model information.</param>
         public LuisService(LuisModelAttribute model)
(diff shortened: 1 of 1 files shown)


willportnoy:
Thank you for the pull request - we decided instead to change the LUIS service query initialization to fix some issues around UTF8 url encoding (in this commit : https://github.com/Microsoft/BotBuilder/commit/8381544055b2e5c3abeb398549d8ebcf0f84193e)

For the other commit, you can simply construct the LuisModelAttribute directly to pass in your model information - it has a public constructor.

evnik:
Well, however I didn't get the point to use `LuisModelAttribute` instead of two dedicated arguments. Actually, I'd like to replace existing constructor with the suggested one, but that breaks compatibility.

PR #409, opened 2016-06-02
Change "port" to "PORT" fix deploy to Heroku

mulyoved:
http://stackoverflow.com/questions/15693192/heroku-node-js-error-web-process-failed-to-bind-to-port-within-60-seconds-of

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/botbuilder/409)


Stevenic:
I need to verify that it doesn't break Azure deployments as all of the Azure samples show "port" lowercased. The samples may need to look for both versions.

mulyoved:
>The samples may need to look for both versions

Updated

Stevenic:
Thanks

PR #422, opened 2016-06-04
Remove deprecation warning in expressjs

k-yle:
`res.send(status, body)` is deprecated in expressjs, replaced it with `res.status(status).send(body)`


Stevenic:
Thanks!

PR #449, opened 2016-06-08
Return normalized string for InitialUpper case

evnik:
Before the fix the joined string was not returned


chrimc62:
I have incorporated this and it will be included in our next release.  Thanks!

evnik:
@chrimc62 could you please explain what's wrong with my PR? If you want me to change something, fix code style or wait for some changes expected to be pushed from your side, then I'm OK with that, just let me know. But I don't like if you reject my contribution and make the same yourself. That kills passion to contribute. Thanks!

PR #495, opened 2016-06-17
Fix restify API

OleksandrBezhan:
In call to `res.status(200)` response returns statusCode (200) instead of response object (this), so there is no chaining.


Stevenic:
thanks..  I had fixed this too but not checked in yet.

PR #506, opened 2016-06-21
Ability to override calculation of the best intent is restored

evnik:
That was pretty useful, but it was missed after implementation of support of multiple LUIS service instances.

In case you want to reject this PR, before that please let me know what should I fix to get it merged, thank you.


willportnoy:
It looks good to me - you might consider a test so that the API doesn't get broken (by me, sorry) in the future.

@msft-shahins can you tell us what branch you want this applied to?

Issue #509, opened 2016-06-21
Can't not run Microsoft.Bot.Sample.SimpleAlarmBot


vnheros:
I tried to run Microsoft.Bot.Sample.SimpleAlarmBot in Bot Framework Emulator, but cannot. I got this error: 500 Internal Server Error
Stack trace:
[...]

PR #511, opened 2016-06-21
Fixed issue #509.

ankitbko:
Issue #509 . 
The problem was making `ILuisService` optional and passing null as default value. `LuisDialog` now accepts `params ILuisService[] services` which was taking null as first item of the array resulting in `NullReferenceException`. Refer to my comment on the issue.
Fixed this by adding a default constructor and removed the default null assignment from other constructor.


ankitbko:
Closing down this to merge it into Develop rather than Master. Is there a guideline where to send pull request to?

PR #512, opened 2016-06-21
Fixed issue #509.

ankitbko:
Issue #509 . 
The problem was making ILuisService optional and passing null as default value. LuisDialog now accepts params ILuisService[] services which was taking null as first item of the array resulting in NullReferenceException. Refer to my comment on the issue.
Fixed this by adding a default constructor and removed the default null assignment from other constructor.


willportnoy:
@msft-shahins this looks fine to me - which branch should we apply it to?

chrimc62:
Thanks!

Issue #552, opened 2016-06-26
Wrong property name.


vvatanabe:
Property name of the Session class is different in botbuilder.d.ts
Look at the following:

current => Session#ISessionState
correct => Session#sessionState

https://github.com/vvatanabe/BotBuilder/blob/master/Node/src/botbuilder.d.ts#L746

v-thgratMSGit:
Closing this issue. Thanks!

Stevenic:
I don't see where this issue is

PR #559, opened 2016-06-27
fix the property name of the Session class

vvatanabe:
Fixed following issue :-)
https://github.com/Microsoft/BotBuilder/issues/552


Stevenic:
thanks

Issue #628, opened 2016-07-09
Basics-logging example middleware not running


TheoRomain:
There is a small bug in the Basics-logging example [app.js](https://github.com/Microsoft/BotBuilder/blob/master/Node/examples/basics-logging/app.js) that causes the middleware to not run.

Botbuilder Version:
3.0.0

Reproduced by:
Running node app.js

Expected behavior:
Use of middleware to display "Logging is now turned on", "Logging is now turned off", "Message received: [message]"

Actual behavior:
bot wrote "UniversalBot.use: no compatible middleware hook found to install." to the console and middleware was not running at all.

Fix:
Changing
`bot.use({
    dialog: function`
to
`bot.use({
    botbuilder: function`
to match the IMiddlewareMap interface allows the example to run as expected.

Stevenic:
good catch. I'll update that. The schema for that changed early last week. forgot to go through the examples.

Stevenic:
fixes checked in

PR #629, opened 2016-07-09
Updated app.js to match IMiddlewareMap hook. Issue: #628

TheoRomain:
This is a pull request for Issue [628](https://github.com/Microsoft/BotBuilder/issues/628)

The original app.js bot.use had "dialog" as the middleware hook in the example usage of "bot.use". According to IMiddlewareMap's definition "dialog" is not a valid hook and thus causes the "Logging" section to never run.

The function defined for "dialog" has "session" and "next" as parameters which corresponds to IMiddlewareMap's "botbuilder" hook.

I updated app.js to change "dialog" to "botbuilder".

The example app.js now runs without error and outputs the logging messages.


Stevenic:
sorry... didn't see that you had also fixed this. there was another example that needed updating as well.

TheoRomain:
No worries. I just noticed I forgot to add a comment about it in the issue. That's my fault. I will keep learning this contributing thing :)

PR #743, opened 2016-07-18
Include ChannelId when creating replay message.

apuyana:
ChannelId was not copied when creating replay message. 
Channel Id can be used in dialogs to modify the answer to better use each channel features.
The service URL was also missing.


mburleigh:
I just spent the morning chasing down the same issue (ServiceUrl not being set in the reply message); until this is merged I will write against local code.

PR #756, opened 2016-07-19
Fix Demo Skype Example

jhernandezme:
SigninCard does not have a `title` function instead should be `text` according to the latest code

https://docs.botframework.com/en-us/node/builder/chat-reference/classes/_botbuilder_d_.signincard.html


Stevenic:
thanks

PR #764, opened 2016-07-20
Support call webhook query parameters

mghimp:
Added new property to Microsoft.Bot.Builder.Calling.Events.IncomingCallEvent:

`public IEnumerable<KeyValuePair<string, string>> IncomingCallParameters { get; set; }`
- additional modifications to support this functionality and backward compatibility of the SDK.

I needed this to be able to configure the incoming call request

### DIFF
--- a/CSharp/Library/Microsoft.Bot.Builder.Calling/CallingBotService.cs
+++ b/CSharp/Library/Microsoft.Bot.Builder.Calling/CallingBotService.cs
@@ -8,6 +8,7 @@ using System.Diagnostics;
 using System.IO;
 using System.Threading.Tasks;
 using System.Threading;
+using System.Collections.Specialized;
 
 namespace Microsoft.Bot.Builder.Calling
 {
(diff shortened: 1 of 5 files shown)


msft-shahins:
Can you elaborate on what is the main application for this feature? Given the registered webhook is fixed for a bot and calling service will not provide different context for a single bot on each incoming call event, does this want to enable multi-bot hosting within the same controller?

mghimp:
The main application of this feature is to allow us to be more operationally efficient, by saving R&D and DevOps costs. 
No, I do not host multiple bots in the same controller. My controller calls always the same (single) bot.
My bot is a conversational bot. The only thing it does is talking to the user. The way my bot is talking to the user is always identical: bot gets user input, then it sends input to our proprietary NLP engine that returns response to the bot, and then bot responds. It is always the same routine.  However, I need this parameter to allow the bot to send the conversation topic to our NLP engine to make an appropriate response. I define different bots on bot platform for each topic and this parameter specifies the topic technically-wise. This allows the bot to send the topic to talk about to NLP engine and thus respond appropriately. The great thing about the implementation of the bot is that it is topic agnostic, i.e. if I add a topic I don't need to modify the bot. The bot is actually used as a "dumb pipe" from user to our NLP engine.  If I don't have the ability to pass parameter to the bot I will have to multiply bot installations, i.e. 1 bot installation for every topic. Then this would require us to hard-wire conversation topic in each bot (R&D costs) and to order additional resources on AWS Elastik Beanstalk (DevOps costs) and thus would be costly for no reason. Since the bot is a "dumb pipe" it does not make any sense to deploy different bot instances for every topic.
Hope I explained the request in a satisfactory manner, but if you have more questions, please do not hesitate to ask.

msft-shahins:
>  I define different bots on bot platform for each topic and this parameter specifies the topic

This implies that you are hosting multiple bots in the same controller (same route) and want to differentiate between them by the different query parameter in the endpoint that you register in the dev portal.

This is a fine approach but you also need to modify BotAuthentication attribute to be able to handle multiple bots, i.e. multiple Microsoft App Id and Microsoft App password and do appropriate authentication based on the query parameter.

mghimp:
I didn't get to BotAuthentication yet because I can't publish my Bot yet because calling feature is still in preview. So while I am still in development mode without being able to publish the bot I left authentication in my TODOs.

msft-shahins:
Closing this because of the age of the pull request. Feel free to send a new pull request based on the new BotAuthentication attribute that support multiple MicrosoftAppId and MicrosoftAppPassword.

mghimp:
Created new PR #1946

PR #885, opened 2016-08-03
Undesired forward resume status in waterfall step

### LINKED ISSUE #883: Cannot call next step with completed status
Using a waterfall strategy with multiple steps, calling `next` (or `skip`) function with  `ResumeReason.completed` resume status, calls next step in waterfall with `ResumeReason.forward` status

### version
nodejs `botbuilder v3.1.1` 
### reproduce
```ts
import * as BotBuilder from 'botbuilder';
const bot = new BotBuilder.UniversalBot();

bot.dialog('bug', [
   resolveDialog,
   printResumed
]);

function resolveDialog(session: BotBuilder.Session, args: any, next: (results?: BotBuilder.IDialogResult<string>) => void) {
  next({
    response: 'hello world',
    resumed: BotBuilder.ResumeReason.completed
  });
}

function printResumed(session: BotBuilder.Session, results: BotBuilder.IDialogResult<string>) {
  console.log(results.resumed);
}
```
### expected behabiour
Executing `session.beginDialog('bug')` should output `0`
### current behabiour
Executing `session.beginDialog('bug')` outputs `4`

jmendiara:
When compiling to typescript, the `ResumeReason.completed` enum evals to `0`
```js
    ResumeReason[ResumeReason["completed"] = 0] = "completed";
```

which is a falsy value. Therefore, any waterfall step that users want to mark as `complete` by using the `next|skip` callback, are automatically marked as `forward` 

Fixes #883

### DIFF
--- a/Node/core/src/dialogs/DialogAction.ts
+++ b/Node/core/src/dialogs/DialogAction.ts
@@ -143,7 +143,7 @@ export function waterfall(steps: IDialogWaterfallStep[]): IDialogHandler<any> {
     return function waterfallAction(s: ses.Session, r: dlg.IDialogResult<any>) {
         var skip = (result?: dlg.IDialogResult<any>) => {
             result = result || <any>{};
-            if (!result.resumed) {
+            if (result.resumed == null) {
                 result.resumed = dlg.ResumeReason.forward;
             }
             waterfallAction(s, result);


Stevenic:
Good catch. I should probably re-evaluate the use of ResumeReason.forward as well

PR #912, opened 2016-08-05
Group conversation sample fix

msimecek:
- ConnectorClient initialization
- await for conversationId
- correct ConversationAccount creation


sustepavkentico:
Hi, 
do you thing these samples for creating the Group chat conversations are still up to date? I am getting "Incorrect conversation creation parameters" message. 
I haven't try only this sample, but almost everything I have found on the internet + tried to solve it with documentation on my own.
**Yet, I was not able to proactively CREATE private group chat conversation and post a message there.**
To use a channel is not a solution for us.

PR #933, opened 2016-08-08
Prevent from server down if request body has nothing

mttrs:
How to reproduce
- Build a bot with 'botbuilder' and 'express'
- Send a request without body
- Server down

Sample bot code:
```javascript
const express = require('express');
const builder = require('botbuilder');
const app = express();

app.get('/', function(req, resp) {
  resp.send('Hello world');
});

const port = process.env.port || 3000;

app.listen(port, function(){
  console.log(`Listening on port ${port}`);
});

var connector = new builder.ChatConnector({
    appId: process.env.MICROSOFT_APP_ID,
    appPassword: process.env.MICROSOFT_APP_PASSWORD
});
var bot = new builder.UniversalBot(connector);
app.post('/api/messages', connector.listen());

bot.dialog('/', function(session) {
    session.send("Hello World");
});
```

Sample request:
`% curl -X POST http://localhost:3978/api/messages`

```
Stack trace
SyntaxError: Unexpected end of JSON input
    at Object.parse (native)
    at IncomingMessage.<anonymous> (botbuilder/lib/bots/ChatConnector.js:43:37)
    at emitNone (events.js:86:13)
    at IncomingMessage.emit (events.js:185:7)
    at endReadableNT (_stream_readable.js:975:12)
    at _combinedTickCallback (internal/process/next_tick.js:74:11)
    at process._tickCallback (internal/process/next_tick.js:98:9)
```

version
```
package.json

  "dependencies": {
    "botbuilder": "^3.2.3",
    "express": "^4.14.0"
  }
```

Reference:
https://nodejs.org/en/docs/guides/anatomy-of-an-http-transaction/#header-request-body

### DIFF
--- a/Node/core/lib/bots/ChatConnector.js
+++ b/Node/core/lib/bots/ChatConnector.js
@@ -35,12 +35,11 @@ var ChatConnector = (function () {
                 _this.verifyBotFramework(req, res);
             }
             else {
-                var requestData = '';
+                var requestData = [];
                 req.on('data', function (chunk) {
-                    requestData += chunk;
-                });
-                req.on('end', function () {
-                    req.body = JSON.parse(requestData);
+                    requestData.push(chunk)
+                }).on('end', function () {
+                    req.body = Buffer.concat(body).toString();
                     _this.verifyBotFramework(req, res);
                 });
             }


Stevenic:
You can add a body-parser and it will use it.  In your change you're not calling JSON.parse() anymore. How is this working?  

Second question is when would you ever not have a body? The ChatConnector is designed to talk to our servers which will always post a body. We can fix the parser exception but your just going to run into other issues.  What exactly are you trying to do?

mttrs:
Using body-parser is better, so I will change it. The current change might be tricky.

It's for preventing from server down. A body paring is done before the auth check, so anyone can shutdown a server with just a single request. It should be avoided and keep a server live. Or it'd not happen?

Stevenic:
We don't want our code to have to require the use of a 'body-parser'... Again... When would you have an empty body in a production bot?

mttrs:
@Stevenic , ok I won't add the body-parser. 
Regarding empty body request, if an application on public, it has IP and the domain. It means that ANYTIME anyone can send any requests to the application. The MS bot's authentication is done AFTER parsing body. It might occur not so much often, but I'd appreciate the botbuilder handles empty body request.

Stevenic:
Seems reasonable...

radustefan95:
I've just run the sample code provided by @mttrs, using v3.9.1 and the issue is still there.
```
SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at IncomingMessage.<anonymous> ([omitted]/node_modules/botbuilder/lib/bots/ChatConnector.js:52:37)
    at emitNone (events.js:105:13)
    at IncomingMessage.emit (events.js:207:7)
    at endReadableNT (_stream_readable.js:1045:12)
    at _combinedTickCallback (internal/process/next_tick.js:138:11)
    at process._tickCallback (internal/process/next_tick.js:180:9)
```
Using a `app.use( bodyParser.json() ); ` with Express/RESTify temporarily solves the issue, but I don't really need a body parser to slow my app down when all that's needed is just a few if clauses...

On a different note, RESTify has it's own bodyParser plugin but I can't get that working. Has anyone managed to do the following?
```
var server = restify.createServer();
server.use( restify.plugins.bodyParser() );
...
```

robinvdvleuten:
It would be nice if it just returns a `400` response. I will open a PR for this :)

PR #988, opened 2016-08-13
changed CallSession.ts to require('./workflow/Prompt')

alexdm:
Changed CallSession.ts to require('./workflow/Prompt') instead of require('./workflow/prompt');

I was trying to deploy my skype calling bot on a PaaS platform on Node 6.3.1. When the app tried to start up, the platform complained with "Error: Cannot find module './workflow/prompt'". 

I changed CallSession.ts import prompt = require('./workflow/prompt') to require('./workflow/Prompt')  and it fixed my problem


Stevenic:
Thanks!

PR #1008, opened 2016-08-16
Extended Attributes.cs with ChoiceFormatOptions.

tomfurrier:
I find strongly typed ChoiceFormats less error-prone than strings.


chrimc62:
Thanks for the pull request, but since these values are typically used in textual templates where the visual look is an important part of the definition, I'm rejecting this pull request.  In a little more detail:
 "{0}. {1}" more clearly communicates a 1. <choice> template than ChoiceFormat.Number + ". " + ChoiceFormat.FieldName.  (And it should probably be ChoiceFormat.FieldDescription if we were going to take the request since Description is the general term for {1}.

Issue #1023, opened 2016-08-17
Node.js ConsoleConnector fails to reset Session.sendingBatch


Thegaram:
The code below sends only one message instead of two.

```js
var connector = new builder.ConsoleConnector().listen();
var bot = new builder.UniversalBot(connector, {
    autoBatchDelay: 0
});

bot.dialog('/', function (session) {
    session.send("First");

    setTimeout(() => { 
        session.send("Second");
    }, 0);
});
```

When we call `Session.send`, the code flow will eventually reach `ConsoleConnector.send`. As this function does not call the provided callback (unlike `ChatConnector.send`), `session.sendingBatch` will not be reset to `false`, and subsequent calls to `Session.send` will be ignored (due to the check in `Session.startBatch`).

Note: without the timeout this issue will not arise, as `Session` does automatic batching inside the time interval defined by `autoBatchDelay` (for which the default is 250ms).

Stevenic:
Looks like you fixed the issue... thanks. Will be in next release.

palmerabollo:
We're also facing this issue. Any plans to release a new nodejs sdk version?

Stevenic:
Should be fixed.

PR #1024, opened 2016-08-17
Use the callback passed to Session.options.onSend everywhere

Thegaram:
The callback passed to this function in `Session.sendBatch` is supposed to reset `session.sendingBatch` to `false`, making it available for further messages. However, in two cases, this callback was not called:

1. In `ConsoleConnector.send` it was simply ignored
2. In `UniversalBot.send` it was not called if an error occured

See #1023 for an example where it caused problems.


Stevenic:
Nice catch

Issue #1031, opened 2016-08-18
EntityRecognizer.parseNumber doesn't work with decimals


oztheozoz:
When Luis.ai detects a decimal number such as 43.4, it successfully parses it as a builtin.number entity. However, Luis adds a space before and after the decimal point like this 43 . 4. EntityRecognizer.parseNumber only parses and returns 43 because of the extra spaces. Either this can be fixed on Luis.ai or by modifying the regex parseNumber relies on.

Tested on botbuilder v 3.2.3

jasondwilliams:
In the response from LUIS, each entity mention includes pointers into the original (pre-tokenized) query string indicating the start and end characters of the entity substring.  Could that be used in parseNumber?

For example:

{
  "query": "hello 4.3 thanks",
  "intents": [
    {
      "intent": "None",
      "score": 0.8677383
    }
  ],
  "entities": [
    {
      "entity": "4 . 3",
      "type": "builtin.number",
      "startIndex": 6,
      "endIndex": 8,
      "score": 0.839945734
    }
  ]
}

The string “4.3” (without spaces) can be recovered with this pseudocode:

startIdx = result[‘entities’][0][‘startIndex’]
endIdx = result[‘entities’][0][‘endIndex’]
entityString = result[‘query’][ startIdx:endIdx ]

oztheozoz:
That is a good work around that I didn't think of. I chose to just modify how ParseNumber works by relaxing the regex a little #1033.

Andrea-Orimoto:
Closing this as it seems you have a work around and a PR going on. Thanks!

oztheozoz:
@Andrea-Orimoto , how will this defect get communicated to the team in charge of Luis?

Andrea-Orimoto:
jason, who replied above, is part of the luis team, and they have captured this issue in their backlog. Thank you!

PR #1033, opened 2016-08-18
Improve numberExp regex

oztheozoz:
When Luis.ai detects a decimal number such as 43.4, it successfully
parses it as a builtin.number entity. However, Luis adds a space before
and after the decimal point like this 43 . 4.
EntityRecognizer.parseNumber only parses and returns 43 because of the
extra spaces.

Link to bug: https://github.com/Microsoft/BotBuilder/issues/1031

### DIFF
--- a/Node/core/lib/dialogs/EntityRecognizer.js
+++ b/Node/core/lib/dialogs/EntityRecognizer.js
@@ -201,7 +201,7 @@ var EntityRecognizer = (function () {
     EntityRecognizer.dateExp = /^\d{4}-\d{2}-\d{2}/i;
     EntityRecognizer.yesExp = /^(1|y|yes|yep|sure|ok|true)/i;
     EntityRecognizer.noExp = /^(2|n|no|nope|not|false)/i;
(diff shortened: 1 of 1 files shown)


Stevenic:
We should actually file this as a bug against LUIS.  Does this still parse 43.4?

oztheozoz:
Yes it does. 






On Mon, Aug 22, 2016 at 11:55 PM -0700, "Steven Ickman" <notifications@github.com> wrote:












We should actually file this as a bug against LUIS.  Does this still parse 43.4?

oztheozoz:
Is a bug going to be opened against LUIS to fix this?

Andrea-Orimoto:
@oztheozoz, yes, it was filed as a bug in LUIS, they were already aware of it. No timeline for a fix though.

oztheozoz:
@Andrea-Orimoto , thx!

oztheozoz:
Any chance this code can be pulled in for the mean time?

Stevenic:
@oztheozoz we're still trying to determine the best way to deal with this... I'm hesitant to bake any logic into the SDK yet as it's more complicated then just decimal points. It's also commas and such.

PR #1209, opened 2016-09-11
Change step's phase from Completed to Ready if field is unknown.

mitkodi:
**FormDialog.MoveToNext** could skip an *Unspecified* field when moving to the next step.
Reproducing the problem:
 1. Create a **FormDialog** with 3 fields (F1, F2, F3).
 2. Let F2 depends on F1, so after setting F1 clear the current value of F2 (in F1 *validate*).
 3. Run the dialog, fill in F1 and F2 and before filling in F3 go back to F1.
After entering a value for F1, the **FormDialog** goes to F3 leaving F2 *Unspecified*. The things could be worse if F3 is active only when F2 is specified. In this case after re-entering F1 the form will be considered as completed.

### DIFF
--- a/CSharp/Library/FormFlow/FormDialog.cs
+++ b/CSharp/Library/FormFlow/FormDialog.cs
@@ -630,6 +630,8 @@ namespace Microsoft.Bot.Builder.FormFlow
                             {
                                 _formState.StepState = null;
                                 _formState.Next = null;
+                                if (_form.Steps[istep].Field.IsUnknown(_state) && _formState.Phase(istep) == StepPhase.Completed)
+                                    _formState.SetPhase(StepPhase.Ready);
                             }
                             if ((_formState.Phase(istep) == StepPhase.Ready || _formState.Phase(istep) == StepPhase.Responding)
                                 && step.Active(_state))


azurecla:
Hi __@mitkodi__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>You've already signed the contribution license agreement. Thanks!</span>
        <span>We will now validate the agreement and then real humans will evaluate your PR.</span>

TTYL, AZPRBOT;

chrimc62:
Thanks for the time and effort to analyze your issue and find a fix, but I'm rejecting this change.  

If a user has entered a value of unspecified, that is a valid value and the step should be seen as completed.  If you want the behavior you are describing, you should do new FieldReflector<someClass>("F1").SetNext((value, state) => new NextStep() { Names=new string[] {"F2"}}).  This is admittedly not as convenient, but you are overriding a reasonable default behavior.

mitkodi:
Many thanks for the suggestion.
Unfortunately, I think that it doesn't work in that case or maybe I miss something. Here the code I use to build the FormDialog:
```
        public static IForm<TestForm> BuildForm()
        {
            var builder = new FormBuilder<TestForm>()
                .Field(new FieldReflector<TestForm>(nameof(TestForm.F1))
                    .SetValidate(async (state, value) =>
                    {
                        state.F2 = null;
                        return new ValidateResult { IsValid = true, Value = value };
                    })
                    .SetNext((value, state) => new NextStep() { Names = new String[] { "F2" } })
                )
                .AddRemainingFields()
                .OnCompletion(async (context, testForm) => await context.PostAsync("Completed"));
            return builder.Build();
        }
```

PR #1283, opened 2016-09-20
Fix missing callback in catch block

palmerabollo:
A callback call was missing in this try/catch block.
The code of this function is a bit weird. It relies on too many try/catch blocks that could be avoided.
In fact, this catch (line 137) looks like a just-in-case catch, I can't figure out why it's needed. You probably could verify it and get rid of it.


azurecla:
Hi __@palmerabollo__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

Stevenic:
This can cause an infinite loop because its the callback that threw the exception. You'd be calling into the callback multiple times.

PR #1320, opened 2016-09-24
[Node SDK] Added a REST/API Direct "Off"line connector

ytechie:
This is a basic example of how you can use the bot builder and create a "direct line" compatible API that allows a bot to be able to use an offline "direct line" when only local network connectivity is allowed, but allows a developer to switch to the public "direct line" API with a simple URL change. This enables offline bots that are better with the cloud.

### DIFF
--- a/Node/core/src/bots/RestfulConnector.ts
+++ b/Node/core/src/bots/RestfulConnector.ts
@@ -1,0 +1,212 @@
+//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license.
+//
+// Microsoft Bot Framework: http://botframework.com
+//
+// Bot Builder SDK Github:
+// https://github.com/Microsoft/BotBuilder
+//
+// Copyright (c) Microsoft Corporation
+// All rights reserved.
+//
+// MIT License:
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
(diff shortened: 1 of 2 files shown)


azurecla:
Hi __@ytechie__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (jayoung). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
        If you are a vendor, DO please sign the electronic contribution license agreement. It will take 2 minutes and there's no faxing! https://cla.azure.com.
    </p>

TTYL, AZPRBOT;

Node/core/src/bots/RestfulConnector.ts
```diff
@@ -0,0 +1,212 @@
...
+interface IOutMessageDictionary {
+    [key: string]: Array<IMessage>;
+};
+
+export class RestfulConnector implements ub.IConnector {
+    private handler: (events: IEvent[], cb?: (err: Error) => void) => void;
+    private rl: readline.ReadLine;
+    private replyCnt = 0;
+
+    private server = restify.createServer();
```
apsamudra: Making server public will allow for bot file to add any other components required.

palmerabollo: I don't think that exposing the internal implementation (restify in this case) is a good idea.

Stevenic:
Give me some time to review this in more detail.

dawnuh:
After review, we decided that we think this will be better in a free standing node module.

stevengum97:
Closing pr.

PR #1383, opened 2016-10-04
[NODE] Fix: Later definition of connectors

### LINKED ISSUE #1382: [Node] Memory Storage used when instantiating Bot w/o connector
Not setting an optional connector parameter when instantiating the bot, but using the `connector` method, causes the bot to use the default storage (memory) instead of the connector one

``` ts
let bot = new BotBuilder.UniversalBot(null, {});
let connector = new BotBuilder.ChatConnector({...});
bot.connector('*', connector);
```

jmendiara:
Currently, a Connector implements `IConnector` and `IBotStorage` (`export class ChatConnector implements ub.IConnector, bs.IBotStorage`) and the `IBotStorage` initialization code is performed only in the constructor.

The FIX mimics the initialization at constructor, but allows defered connector setting:  only the first `IBotStorage` implementation from the first `Connector` provided is used as the provided storage. Next calls to `connector` only adds the `IConnector` functionality, discarding the *mixed* `IBotStorage` part

Fixes #1382

### DIFF
--- a/Node/core/src/bots/UniversalBot.ts
+++ b/Node/core/src/bots/UniversalBot.ts
@@ -108,12 +108,6 @@ export class UniversalBot extends events.EventEmitter {
 
         if (connector) {
             this.connector(consts.defaultConnector, connector);
-            var asStorage: bs.IBotStorage = <any>connector;
-            if (!this.settings.storage && 
-                typeof asStorage.getData === 'function' &&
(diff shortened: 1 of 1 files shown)


azurecla:
    <span>
    </span>

TTYL, AZPRBOT;

Stevenic:
I made the same basic fix so closing.

PR #1402, opened 2016-10-07
[Node] Propagate types in endDialogWithResult

jmendiara:
`IDialogResult<T>` interface has a template syntax, so `endDialogWithResult` which takes that interface as argument, should propagate the template

### DIFF
--- a/Node/core/lib/botbuilder.d.ts
+++ b/Node/core/lib/botbuilder.d.ts
@@ -1,4 +1,4 @@
-﻿//=============================================================================
+//=============================================================================
 //
 // INTERFACES
 //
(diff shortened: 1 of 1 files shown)


azurecla:
    <span>
    </span>

TTYL, AZPRBOT;

Stevenic:
So that's not actually going to achieve anything. Can you show me an example where that would matter?

jmendiara:
Yeah, you are right. This does not matter at all in userland. I was just looking for consistency within your definitions

https://github.com/Microsoft/BotBuilder/blob/master/Node/core/src/botbuilder.d.ts#L1133
```ts
beginDialog<T>(id: string, args?: T): Session;
```

https://github.com/Microsoft/BotBuilder/blob/master/Node/core/src/botbuilder.d.ts#L1141
```ts
 replaceDialog<T>(id: string, args?: T): Session;
```

Stevenic:
Fair enough

PR #1403, opened 2016-10-07
[Node] Definitions for IDialogResult.resumed

jmendiara:
The current implementation [makes a guard over not defining the `resumed` property](https://github.com/Microsoft/BotBuilder/blob/e8c5aaaa273f6265fd983e5ded2f3b147a9867b6/Node/core/src/Session.ts#L297-L299), so it should be optional in TS to avoid generating verbose code

before
```ts
session.endDialogWithResult({
  resumed: BotBuilder.ResumeReason.completed,
  response: true
});
```

after
```ts
session.endDialogWithResult({
  response: true
});
```

### DIFF
--- a/Node/core/src/botbuilder.d.ts
+++ b/Node/core/src/botbuilder.d.ts
@@ -385,8 +385,8 @@ export interface IDialogState {
   * Results returned by a child dialog to its parent via a call to session.endDialog(). 
   */
 export interface IDialogResult<T> {
-    /** The reason why the current dialog is being resumed. */
-    resumed: ResumeReason;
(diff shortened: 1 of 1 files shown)


azurecla:
    <span>
    </span>

TTYL, AZPRBOT;

Stevenic:
That's fair. I went back and forth over that. Depends on which way you're optimizing for. It actually is required but in practice no one ever fills it in so I'm ok marking it optional.

jmendiara:
> but in practice no one ever fills it 

FYI: As we code our bot in Typescript, we MUST fill it to get transpilation working

Stevenic:
No I agree... I can see how its annoying in typescript.

PR #1533, opened 2016-10-27
Authorise MSA self-issued tokens in the node SDK (as in the .NET SDK)…

elliatab:
The C# SDK allows to use self-issues tokens, not only for the emulator but for any channel as long as the AppId of the Bot is matching the one from the token:
https://github.com/Microsoft/BotBuilder/blob/5cf71c742f27d89e9dbc4076f850122bd6edac11/CSharp/Library/Microsoft.Bot.Connector/BotAuthentication.cs#L45

Additionally, I think the node SDK currently allows any MSA-issued token to communicate with the bot as long as the channelId field contains "emulator". I think it may be a security issue. At least the SDK should also check the appId for the emulator channel.

### DIFF
--- a/Node/core/lib/bots/ChatConnector.js
+++ b/Node/core/lib/bots/ChatConnector.js
@@ -63,7 +63,13 @@ var ChatConnector = (function () {
             var decoded = jwt.decode(token, { complete: true });
             var verifyOptions;
             var openIdMetadata;
-            if (isEmulator && decoded.payload.iss == this.settings.endpoint.msaIssuer) {
+            if (decoded.payload.iss == this.settings.endpoint.msaIssuer) {
+                if(decoded.payload.appid !== this.settings.appId) {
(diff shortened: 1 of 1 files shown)


azurecla:
Hi __@elliatab__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

elliatab:
to be consistent with the C# SDK I would prefer to avoid the check for the isEmulator flag. I think this check is only useful for the case where all security is disabled in development:
https://github.com/elliatab/BotBuilder/blob/fba9cbf47bde35fff5889e37c4bbdc75275d5d35/Node/core/lib/bots/ChatConnector.js#L109

elliatab:
@dandriscoll any update on this? Do you really want to keep the check on isEmulator flag?

dandriscoll:
Makes sense. Thanks!

PR #1571, opened 2016-11-02
Fix loop indexes are being meshed up in the ConsoleConnector

### LINKED ISSUE #1570: [Node.js] ConsoleConnector is behaving wrong when a bot sends attachments
The ConsoleConnector is behaving wrong in the following cases:
* Case 1:
  * Send a message with attachments
  * Send a Prompts.choice message with listStyle = button
  * The ConsoleConnector is not printing the choices (which are actually cards)
* Case 2:
  * Send a message with only text
  * Send a message with attachments
  * Send a Prompts.choice message with listStyle = button
  * The ConsoleConnector enters a loop printing the choices forever

JoseAntonioRodriguez:
Fixes #1570

The ConsoleConnector is declaring the same variable `i` in two nested loops, which mades the inner loop to modify the value of the outer loop index.
Some of the consequences of this bug are described in #1570.

This could be solved using `let` instead of `var` but I've seen that you are not using `let` in the framework, so I've renamed the inner loop index from `i` to `j`.

### DIFF
--- a/Node/core/src/bots/ConsoleConnector.ts
+++ b/Node/core/src/bots/ConsoleConnector.ts
@@ -40,7 +40,7 @@ export class ConsoleConnector implements ub.IConnector {
     private handler: (events: IEvent[], cb?: (err: Error) => void) => void;
     private rl: readline.ReadLine;
     private replyCnt = 0;
-    
+
     public listen(): this {
         this.rl = readline.createInterface({ input: process.stdin, output: process.stdout, terminal: false });
         this.rl.on('line', (line: string) => {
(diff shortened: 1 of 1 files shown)


azurecla:
Hi __@JoseAntonioRodriguez__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

Stevenic:
great catch. thanks!

palmerabollo:
let for the win 😀

Issue #1597, opened 2016-11-07
Facebook is missing as a trusted hostname


vossccp:
Beginning wit 3.3.1, the connector is checking against a list of trusted hostnames. As of 3.3.1 this only includes call to state.botframework.com. Calls to facebook.botframework.com get rejected.

Source MicrosoftAppCredentials#TrustedUri()

tomlm:
The deal with the trusted service urls is that it is not an exhaustive list.  We regularly add new channels and endpoints for channels.  Whenever an incoming activity comes in and is a trusted service because it is signed with the appropriate Jwt token, we add the url to our trusted list, with a 24 hour expiration.

As long as you are getting signed data from that url you should be good.  You will only end up trouble if:
 you are proactively sending a message to a channel 
AND 
it has not received any signed activities in the last 24 hours.

vossccp:
I believe, the issue we encounter is related to the way, we deal with messages from the Bot Framework: When we receive a message, we route the message into a queue; we have different endpoints (each of them representing specialised bots) which then process the message.

Adding the URLs when the message arrives, will not work in such distributed scenarios because the instances dont share the same memory space.

I think, the framework should generally be able to support this, so I suggest reconsidering the setup

PR #1598, opened 2016-11-07
facebook is missing as a trusted hostname

vossccp:
Connector calls to facebook.botframework.com are Forbidden because the the url is not part of the list of trusted hostnames. See #1597


azurecla:
Hi __@vossccp__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

msft-shahins:
Trusted urls will be added to trusted host names by BotAtuhentication only for requests that have a valid JwtToken. If you want to trust a service url in your bot you can call [MicrosoftAppCredentails.TrustServiceUrl](https://github.com/Microsoft/BotBuilder/blob/master/CSharp/Library/Microsoft.Bot.Connector/MicrosoftAppCredentials.cs#L56) to trust that host name, e.g. facebook.botframework.com.

PR #1614, opened 2016-11-09
Allow capital case authorization header

### LINKED ISSUE #1613: Authentication fails if authorization header starts with an upper case letter
The Microsoft Bot Framework server sends the authorization header starting with a lower case letter and the Node.js bot code does the lookup based on lower-case key. However, there might be a setup where there is a gateway-style component between the Bot Framework and the actual bot implementation. If the gateway does a normalization for the HTTP headers (converts the header from `authorization` -> `Authorization`), the header lookup in the Node.js code fails.

vjrantal:
Fixes #1613.

### DIFF
--- a/Node/core/lib/bots/ChatConnector.js
+++ b/Node/core/lib/bots/ChatConnector.js
@@ -51,9 +51,10 @@ var ChatConnector = (function () {
         var _this = this;
         var token;
         var isEmulator = req.body['channelId'] === 'emulator';
-        if (req.headers && req.headers.hasOwnProperty('authorization')) {
-            var auth = req.headers['authorization'].trim().split(' ');
-            ;
+        var authorizationHeaderValue = req.headers &&
+            req.headers['authorization'] || req.headers['Authorization'];
+        if (authorizationHeaderValue) {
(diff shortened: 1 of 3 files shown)


azurecla:
Hi __@vjrantal__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (ville.rantala). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
        If you are a vendor, DO please sign the electronic contribution license agreement. It will take 2 minutes and there's no faxing! https://cla.azure.com.
    </p>

TTYL, AZPRBOT;

vjrantal:
@Stevenic I created this against PR against `master`, but noticed that there is also branch `develop`. Let me know, if that would be a better target branch.

Stevenic:
Thanks I've made a slightly different fix that doesn't depend on knowing operator precedence:

```
var authHeaderValue = req.headers ? req.headers['authorization'] || req.headers['Authorization'] : null;
```
I'll also add your unit tests.

PR #1640, opened 2016-11-11
Fixed "Parameter count missmatch" exception for setup of QueryAsync in LuisTestBase.

grbinho:
Hi,

I was trying to write tests for a bot I'm building. I was looking for an inspiration in your tests. Specifically [`AlarmBotTests`](https://github.com/Microsoft/BotBuilder/blob/master/CSharp/Tests/Microsoft.Bot.Sample.Tests/AlarmBotTests.cs).

In my tests I was running into issues with "Parameter count mismatch" exceptions when invoking tests classes, so I looked at you tests more closely. 
As far as I can tell, you are not actually using mocked LUIS services even though they are set up in the test. `AlarmModule` registers actual LuisService implementation. But that is another issue.

When using actual mocked service setup should be like this, otherwise it will fails with "Parameter count missmatch" exception.

### DIFF
--- a/CSharp/Tests/Microsoft.Bot.Builder.Tests/LuisTests.cs
+++ b/CSharp/Tests/Microsoft.Bot.Builder.Tests/LuisTests.cs
@@ -1,15 +1,15 @@
-﻿// 
+﻿//
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license.
-// 
+//
 // Microsoft Bot Framework: http://botframework.com
-// 
+//
 // Bot Builder SDK Github:
 // https://github.com/Microsoft/BotBuilder
-// 
+//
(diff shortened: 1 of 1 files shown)


azurecla:
Hi __@grbinho__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

willportnoy:
@grbinho thank you for the contribution - I fixed the alarm bot tests in this commit: https://github.com/Microsoft/BotBuilder/commit/d7ea6cfcfcfa515210c5f7771ca01f12ab427ea8

As you pointed out, it wasn't even using the mock service, so I added the mock verification at the end of the test as well as fixed a bunch of other issues.

PR #1807, opened 2016-12-07
Populate reply message with richer details in the node SDK

mankityau:
The message now send back to the web connector with the Node SDK is missing some fields compare to the C# SDK. For example, replyToID and etc. 

I looked into the C# library' createReply method (https://github.com/Microsoft/BotBuilder/blob/master/CSharp/Library/Microsoft.Bot.Connector/ActivityEx.cs) and made some adjustment. So that the two SDK behave more like each other.

### DIFF
--- a/Node/core/lib/bots/ChatConnector.js
+++ b/Node/core/lib/bots/ChatConnector.js
@@ -356,11 +356,8 @@ var ChatConnector = (function () {
     };
     ChatConnector.prototype.postMessage = function (msg, cb) {
         logger.info(address, 'ChatConnector: sending message.');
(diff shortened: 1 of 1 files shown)


azurecla:
Hi __@mankityau__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

mankityau:
Hi, any news about this?

Stevenic:
The added fields that the C# SDK returns aren't needed. The Node SDK is more compliant with the spec so I'd prefer not to do this.

PR #1879, opened 2016-12-18
fix tsc errors TS2399, TS2400 in eventMiddleware

sebsylvester:
Compiling with tsc led to the following errors:
bots/UniversalBot.ts(470,13): error TS2399: Duplicate identifier '_this'. Compiler uses variable declaration '_this' to capture 'this' reference.
bots/UniversalBot.ts(473,17): error TS2400: Expression resolves to variable declaration '_this' that compiler uses to capture 'this' reference.
bots/UniversalBot.ts(477,17): error TS2400: Expression resolves to variable declaration '_this' that compiler uses to capture 'this' reference.

Caching the 'this' pointer wasn't necessary, calling ```next.apply(this)``` has the same effect.

### DIFF
--- a/Node/core/src/bots/UniversalBot.ts
+++ b/Node/core/src/bots/UniversalBot.ts
@@ -467,17 +467,16 @@ export class UniversalBot extends Library {
 
     private eventMiddleware(event: IEvent, middleware: IEventMiddleware[], done: Function, error?: (err: Error) => void): void {
         var i = -1;
-        var _this = this;
         function next() {
             if (++i < middleware.length) {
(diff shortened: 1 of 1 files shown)


azurecla:
Hi __@sebsylvester__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

Stevenic:
That's a good suggestion with the `next.apply(this)` is a good one but I already fixed this issue. Let me think about adopting your approach.

Issue #2312, opened 2017-02-20
Locale issue with cardAction.imBack


strangecyan:
I'm creating a card that has a button which has a different imBack value than the title of the button.

I need to be able to localise that message (in this case replacing "Can I book a holiday?" with "Can I book a vacation?"). The title is being correctly localised but the message is being sent as 'home_absence_message_book'. Is this how the framework currently works and if it is, can I raise a feature request to change it?

`builder.CardAction.imBack(session, 'home_absence_message_book', 'home_absence_button_book')`

Stevenic:
So it's actually by design that the payload isn't being localized.  If you want to localize the payload you can do so pretty easily:

```
builder.CardAction.imBack(session, session.gettext('home_absence_message_book'), 'home_absence_button_book')
```

Stevenic:
Closing this for now. Please feel free to add further comments.

PR #2313, opened 2017-02-20
Localisation for cardAction imBack value

strangecyan:
As referenced in #2312, I had a situation where I needed to localise the imBack value. This is a possible implementation that allows for that imBack value to be localised but does not try to attempt to localise any of the other cardAction types in case there are some issues with urls, etc.


azurecla:
Hi __@strangecyan__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

Stevenic:
See my answer in your original post. We already have a fairly simple means of localizing the payload of an action.

Issue #2387, opened 2017-03-03
Stop requiring Activity for simple convenience methods


brandonh-msft:
In many cases the dialog flow utilizes `IMessageActivity` yet we have many nice methods off `Activity` that are helpful on `IMessageActivity` types (eg: `GetStateClient`)

Instead of relegating these nice methods to a heavy object, let's make them extension methods.

PR coming! :)

brandonh-msft:
Fixed by #2672

PR #2457, opened 2017-03-17
[C# SDK] Api feedback improvements

brandonh-msft:
This PR contains a number of small, but impactful, changes to the API to make it more digestable. These feedback points came out of a hack with a partner.

1. In order to use `GetStateClient()` today you **must** use the full-blown `Activity` object. This is less desirable than using one of the interfaces from which `Activity` inherits (which has all the necessary information to perform the same work). So, I changed those methods to be extensions off the interface. Addresses #2387 
1. It's confusing that `IBotDataBag` has a `Set*Value*` method but only a `Get` method (one would reasonably expect `Get*Value*` here). Fixed and marked `Get` obsolete with a warning.
1. Added a method `GetValueOrDefault` on `IBotDataBag` so users can retrieve a value from a data bag or, optionally, get back either `default(T)` or a default of their choosing. Avoids much `if/else` boilerplate when retrieving values.
2. The `Forward` method on `IDialogContext` required a `cancellationToken` parameter, but other `IDialogContext` methods treat this as optional. Unified this.
3. During our discussions during the hack w/ the BF team, we found that `ETag` should be set to `"*"` when updating data sets in the state client so set this as the default value for `ETag`.
4. Checked throughout the framework for the optional status of the `cancellationToken` elsewhere. Unified it with a default value of `default(CancellationToken)`
5. `PromptDialog.Choice` used different verbiage for labeling the `CardAction` `title` and `value` parameters. I unified this terminology and also added an overload that made more sense; one that takes in an `IEnumerable<KeyValuePair<string,T>>` which is a more obvious descriptor of the purpose of the parameter.
6. Often developers end up with a `NullReferenceException` at runtime because they have done something like this:
```
var card = new HeroCard();
card.Buttons.Add(new CardAction("buttonText"));
```
This blows up because `HeroCard`'s `Buttons` property is initialized to `null`. Fixed this throughout our Card objects.

9. `CardAction` can be constructed and sent out without the `Type` property defined. This isn't valid and results in no response sent to the user (and an exception within the bot). Instead, default `Type` to something; I chose `ImBack`.
10. `Activity` can be created (eg: `new Activity()`) and will immediately blow up if a developer tries to `.Attachments.Add(new HeroCard().ToAttachment())` because the `Attachments` property is null. Instead of only initializing it to blank during `CreateReply()`, just ensure it's blank and `.Add`-able when defaultly constructed.
11. The Action type `Invoke` is not on the `ActionTypes` constants class. This is a valid Action Type, so it should be present.
12. Despite the XML doco, `IActivity.GetChannelData<T>()` does not return `default(T)` if it's not able to coerce `.ChannelData` in to type `T`. Rather, it throws an exception. Added `IActivity.TryGetValue<T>(out T instance)` to the set of extension methods available on `IActivity`.

### DIFF
--- a/CSharp/Library/Microsoft.Bot.Builder/ConnectorEx/BotData.cs
+++ b/CSharp/Library/Microsoft.Bot.Builder/ConnectorEx/BotData.cs
@@ -38,10 +38,8 @@ using Newtonsoft.Json.Linq;
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
-using System.Configuration;
 using System.IO;
 using System.IO.Compression;
-using System.Linq;
 using System.Net;
 using System.Threading;
 using System.Threading.Tasks;
@@ -65,7 +63,7 @@ namespace Microsoft.Bot.Builder.Dialogs.Internals
         /// <param name="botStoreType"> The bot store type.</param>
         /// <param name="cancellationToken"> The cancellation token.</param>
         /// <returns>Bot record that is stored for this key, or "empty" bot record ready to be stored</returns>
-        Task<T> LoadAsync(IAddress key, BotStoreType botStoreType, CancellationToken cancellationToken);
+        Task<T> LoadAsync(IAddress key, BotStoreType botStoreType, CancellationToken cancellationToken = default(CancellationToken));
 
         /// <summary>
         /// Save a BotData using the ETag.
@@ -80,8 +78,8 @@ namespace Microsoft.Bot.Builder.Dialogs.Internals
         /// <param name="data"> The data that should be saved.</param>
         /// <param name="cancellationToken"> The cancellation token.</param>
         /// <returns>throw HttpException(HttpStatusCode.PreconditionFailed) if update fails</returns>
-        Task SaveAsync(IAddress key, BotStoreType botStoreType, T data, CancellationToken cancellationToken);
-        Task<bool> FlushAsync(IAddress key, CancellationToken cancellationToken);
+        Task SaveAsync(IAddress key, BotStoreType botStoreType, T data, CancellationToken cancellationToken = default(CancellationToken));
+        Task<bool> FlushAsync(IAddress key, CancellationToken cancellationToken = default(CancellationToken));
     }
 
     /// <summary>
@@ -144,7 +142,7 @@ namespace Microsoft.Bot.Builder.Dialogs.Internals
             }
         }
 
-        Task<bool> IBotDataStore<BotData>.FlushAsync(IAddress key, CancellationToken cancellationToken)
+        Task<bool> IBotDataStore<BotData>.FlushAsync(IAddress key, CancellationToken cancellationToken )
         {
             // Everything is saved. Flush is no-op
             return Task.FromResult(true);
@@ -222,7 +220,7 @@ namespace Microsoft.Bot.Builder.Dialogs.Internals
             return botData;
         }
 
-        async Task IBotDataStore<BotData>.SaveAsync(IAddress key, BotStoreType botStoreType, BotData botData, CancellationToken cancellationToken)
+        async Task IBotDataStore<BotData>.SaveAsync(IAddress key, BotStoreType botStoreType, BotData botData, CancellationToken cancellationToken )
         {
             switch (botStoreType)
             {
@@ -240,7 +238,7 @@ namespace Microsoft.Bot.Builder.Dialogs.Internals
             }
         }
 
-        Task<bool> IBotDataStore<BotData>.FlushAsync(IAddress key, CancellationToken cancellationToken)
+        Task<bool> IBotDataStore<BotData>.FlushAsync(IAddress key, CancellationToken cancellationToken )
         {
             // Everything is saved. Flush is no-op
             return Task.FromResult(true);
@@ -287,7 +285,7 @@ namespace Microsoft.Bot.Builder.Dialogs.Internals
             public BotData BotUserData { set; get; }
         }
 
-        async Task<bool> IBotDataStore<BotData>.FlushAsync(IAddress key, CancellationToken cancellationToken)
+        async Task<bool> IBotDataStore<BotData>.FlushAsync(IAddress key, CancellationToken cancellationToken )
         {
             CacheEntry entry;
             if (cache.TryGetValue(key, out entry))
@@ -306,7 +304,7 @@ namespace Microsoft.Bot.Builder.Dialogs.Internals
             }
         }
 
-        async Task<BotData> IBotDataStore<BotData>.LoadAsync(IAddress key, BotStoreType botStoreType, CancellationToken cancellationToken)
+        async Task<BotData> IBotDataStore<BotData>.LoadAsync(IAddress key, BotStoreType botStoreType, CancellationToken cancellationToken )
         {
             CacheEntry cacheEntry;
             BotData value = null;
@@ -351,7 +349,7 @@ namespace Microsoft.Bot.Builder.Dialogs.Internals
             return value;
         }
 
-        async Task IBotDataStore<BotData>.SaveAsync(IAddress key, BotStoreType botStoreType, BotData value, CancellationToken cancellationToken)
+        async Task IBotDataStore<BotData>.SaveAsync(IAddress key, BotStoreType botStoreType, BotData value, CancellationToken cancellationToken )
         {
             CacheEntry entry;
             if (!cache.TryGetValue(key, out entry))
@@ -363,7 +361,7 @@ namespace Microsoft.Bot.Builder.Dialogs.Internals
             SetCachedValue(entry, botStoreType, value);
         }
 
-        private async Task<BotData> LoadFromInnerAndCache(CacheEntry cacheEntry, BotStoreType botStoreType, IAddress key, CancellationToken token)
+        private async Task<BotData> LoadFromInnerAndCache(CacheEntry cacheEntry, BotStoreType botStoreType, IAddress key, CancellationToken token = default(CancellationToken))
         {
             var value = await inner.LoadAsync(key, botStoreType, token);
 
@@ -398,7 +396,7 @@ namespace Microsoft.Bot.Builder.Dialogs.Internals
             }
         }
 
-        private async Task Save(IAddress key, CacheEntry entry, CancellationToken cancellationToken)
+        private async Task Save(IAddress key, CacheEntry entry, CancellationToken cancellationToken = default(CancellationToken))
         {
             switch (this.dataConsistencyPolicy)
             {
@@ -449,7 +447,7 @@ namespace Microsoft.Bot.Builder.Dialogs.Internals
     {
         private readonly IBotData inner;
         private readonly IDialogTaskManager dialogTaskManager;
-        
+
         public DialogTaskManagerBotDataLoader(IBotData inner, IDialogTaskManager dialogTaskManager)
         {
             SetField.NotNull(out this.inner, nameof(inner), inner);
@@ -459,19 +457,19 @@ namespace Microsoft.Bot.Builder.Dialogs.Internals
         public IBotDataBag UserData { get { return inner.UserData; } }
         public IBotDataBag ConversationData { get { return inner.ConversationData; } }
         public IBotDataBag PrivateConversationData { get { return inner.PrivateConversationData; } }
-        public async Task LoadAsync(CancellationToken token)
+        public async Task LoadAsync(CancellationToken token = default(CancellationToken))
         {
             await this.inner.LoadAsync(token);
             await this.dialogTaskManager.LoadDialogTasks(token);
         }
 
-        public async Task FlushAsync(CancellationToken token)
+        public async Task FlushAsync(CancellationToken token = default(CancellationToken))
         {
             await this.dialogTaskManager.FlushDialogTasks(token);
             await this.inner.FlushAsync(token);
         }
     }
-    
+
     public abstract class BotDataBase<T> : IBotData
     {
         protected readonly IBotDataStore<BotData> botDataStore;
@@ -489,7 +487,7 @@ namespace Microsoft.Bot.Builder.Dialogs.Internals
         protected abstract T MakeData();
         protected abstract IBotDataBag WrapData(T data);
 
-        public async Task LoadAsync(CancellationToken cancellationToken)
+        public async Task LoadAsync(CancellationToken cancellationToken = default(CancellationToken))
         {
             var conversationTask = LoadData(BotStoreType.BotConversationData, cancellationToken);
             var privateConversationTask = LoadData(BotStoreType.BotPrivateConversationData, cancellationToken);
@@ -500,7 +498,7 @@ namespace Microsoft.Bot.Builder.Dialogs.Internals
             this.userData = await userTask;
         }
 
-        public async Task FlushAsync(CancellationToken cancellationToken)
+        public async Task FlushAsync(CancellationToken cancellationToken = default(CancellationToken))
         {
             await this.botDataStore.FlushAsync(botDataKey, cancellationToken);
         }
@@ -532,7 +530,7 @@ namespace Microsoft.Bot.Builder.Dialogs.Internals
             }
         }
 
-        private async Task<IBotDataBag> LoadData(BotStoreType botStoreType, CancellationToken cancellationToken)
+        private async Task<IBotDataBag> LoadData(BotStoreType botStoreType, CancellationToken cancellationToken = default(CancellationToken))
         {
             var botData = await this.botDataStore.LoadAsync(botDataKey, botStoreType, cancellationToken);
             if (botData?.Data == null)
--- a/CSharp/Library/Microsoft.Bot.Builder/ConnectorEx/BotToUser.cs
+++ b/CSharp/Library/Microsoft.Bot.Builder/ConnectorEx/BotToUser.cs
@@ -31,6 +31,7 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 //
 
+using Microsoft.Bot.Builder.ConnectorEx;
 using System;
 using System.Collections.Generic;
 using System.IO;
@@ -38,7 +39,6 @@ using System.Linq;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.Bot.Builder.ConnectorEx;
 using Microsoft.Bot.Builder.Internals.Fibers;
 using Microsoft.Bot.Connector;
 
@@ -78,7 +78,7 @@ namespace Microsoft.Bot.Builder.Dialogs.Internals
             return toBotActivity.CreateReply();
         }
 
-        Task IBotToUser.PostAsync(IMessageActivity message, CancellationToken cancellationToken)
+        Task IBotToUser.PostAsync(IMessageActivity message, CancellationToken cancellationToken )
         {
             return Task.CompletedTask;
         }
@@ -100,7 +100,7 @@ namespace Microsoft.Bot.Builder.Dialogs.Internals
             return toBotActivity.CreateReply();
         }
 
-        async Task IBotToUser.PostAsync(IMessageActivity message, CancellationToken cancellationToken)
+        async Task IBotToUser.PostAsync(IMessageActivity message, CancellationToken cancellationToken )
         {
             await this.client.Conversations.ReplyToActivityAsync((Activity)message, cancellationToken);
         }
@@ -286,7 +286,7 @@ namespace Microsoft.Bot.Builder.Dialogs.Internals
             return toBotActivity.CreateReply();
         }
 
-        async Task IBotToUser.PostAsync(IMessageActivity message, CancellationToken cancellationToken)
+        async Task IBotToUser.PostAsync(IMessageActivity message, CancellationToken cancellationToken )
         {
             this.queue.Enqueue(message);
         }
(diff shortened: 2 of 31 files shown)


CSharp/Library/Microsoft.Bot.Builder/ConnectorEx/BotData.cs
```diff
@@ -65,7 +63,7 @@ public interface IBotDataStore<T>
         /// <param name="botStoreType"> The bot store type.</param>
         /// <param name="cancellationToken"> The cancellation token.</param>
         /// <returns>Bot record that is stored for this key, or "empty" bot record ready to be stored</returns>
-        Task<T> LoadAsync(IAddress key, BotStoreType botStoreType, CancellationToken cancellationToken);
+        Task<T> LoadAsync(IAddress key, BotStoreType botStoreType, CancellationToken cancellationToken = default(CancellationToken));
```
willportnoy: > = default(CancellationToken) [](start = 102, length = 29)

default arguments on interfaces are odd because they're not applied consistently to the derived class.

in addition, making CancellationToken optional encourages people to break cooperative cancellation by not passing the token.

CSharp/Library/Microsoft.Bot.Builder/ConnectorEx/BotToUser.cs
```diff
@@ -100,7 +100,7 @@ IMessageActivity IBotToUser.MakeMessage()
             return toBotActivity.CreateReply();
         }
 
-        async Task IBotToUser.PostAsync(IMessageActivity message, CancellationToken cancellationToken)
+        async Task IBotToUser.PostAsync(IMessageActivity message, CancellationToken cancellationToken )
```
willportnoy: >  [](start = 101, length = 1)

?

CSharp/Library/Microsoft.Bot.Builder/ConnectorEx/IBotDataBag.cs
```diff
@@ -89,14 +89,27 @@ public interface IBotDataBag
     /// </summary>
     public static partial class Extensions
     {
+        [System.Obsolete(@"Use GetValue<T> instead", false)]
+        public static T Get<T>(this IBotDataBag bag, string key)
```
willportnoy: >Get [](start = 24, length = 3)

good to fix naming consistency, thank you :)

CSharp/Library/Microsoft.Bot.Builder/Dialogs/PromptDialog.cs
```diff
@@ -362,7 +362,7 @@ public static void Text(IDialogContext context, ResumeAfter<string> resume, stri
         /// <param name="patterns">Yes and no alternatives for matching input where first dimension is either <see cref="PromptConfirm.Yes"/> or <see cref="PromptConfirm.No"/> and the arrays are alternative strings to match.</param>
         public static void Confirm(IDialogContext context, ResumeAfter<bool> resume, string prompt, string retry = null, int attempts = 3, PromptStyle promptStyle = PromptStyle.Auto, string[] options = null, string[][] patterns = null)
         {
-            Confirm(context, resume, new PromptOptions<string>(prompt, retry, attempts: attempts, options: options ?? PromptConfirm.Options, promptStyler: new PromptStyler(promptStyle: promptStyle)), patterns);
+            Confirm(context, resume, new PromptOptions<string>(prompt, values: default(IReadOnlyList<string>), retry: retry, attempts: attempts, promptStyler: new PromptStyler(promptStyle: promptStyle)));
```
willportnoy: >values: default(IReadOnlyList<string>) [](start = 71, length = 38)

isn't this the default value?

CSharp/Library/Microsoft.Bot.Connector.Shared/ActionTypes.cs
```diff
@@ -47,5 +47,10 @@ public class ActionTypes
         /// Signin button
         /// </summary>
         public const string Signin = "signin";
+
+        /// <summary>
+        /// Invoke a command back to the bot
+        /// </summary>
+        public const string Invoke = "invoke";
```
willportnoy: >invoke [](start = 38, length = 6)

I thought this was an activity type, not an action type.

CSharp/Library/Microsoft.Bot.Connector.Shared/ActivityEx.cs
```diff
@@ -149,17 +147,57 @@ public Activity CreateReply(string text = null, string locale = null)
...
+
+            if (String.Equals(type, ActivityTypes.DeleteUserData, StringComparison.OrdinalIgnoreCase))
+                return ActivityTypes.DeleteUserData;
+
+            if (String.Equals(type, ActivityTypes.Typing, StringComparison.OrdinalIgnoreCase))
+                return ActivityTypes.Typing;
+
+            if (String.Equals(type, ActivityTypes.Ping, StringComparison.OrdinalIgnoreCase))
+                return ActivityTypes.Ping;
+
```
willportnoy: you might consider using a new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase) to make your canonicalization intent clear

CSharp/Library/Microsoft.Bot.Connector.Shared/ActivityEx.cs
```diff
@@ -149,17 +147,57 @@ public Activity CreateReply(string text = null, string locale = null)
...
+            return $"{Char.ToLower(type[0])}{type.Substring(1)}";
+        }
+
+        public bool HasContent()
+        {
+            // hit the extension method
```
willportnoy: >// hit the extension method [](start = 12, length = 27)

isn't the extension method available without this helper method?

CSharp/Library/Microsoft.Bot.Connector.Shared/ActivityEx.cs
```diff
@@ -149,17 +147,57 @@ public Activity CreateReply(string text = null, string locale = null)
...
+        /// <summary>
         /// Get StateClient appropriate for this activity
         /// </summary>
         /// <param name="credentials">credentials for bot to access state api</param>
         /// <param name="serviceUrl">alternate serviceurl to use for state service</param>
         /// <param name="handlers"></param>
+        /// <param name="activity"></param>
         /// <returns></returns>
-        public StateClient GetStateClient(MicrosoftAppCredentials credentials, string serviceUrl = null, params DelegatingHandler[] handlers)
+        public static StateClient GetStateClient(this IActivity activity, MicrosoftAppCredentials credentials, string serviceUrl = null, params DelegatingHandler[] handlers)
```
willportnoy: >GetStateClient [](start = 34, length = 14)

this seems like a good idea, thank you :)

CSharp/Library/Microsoft.Bot.Connector.Shared/ActivityEx.cs
```diff
@@ -174,28 +212,82 @@ public StateClient GetStateClient(MicrosoftAppCredentials credentials, string se
...
+            {
+                if (activity.ChannelData == null)
+                {
+                    return false;
+                }
+
+                instance = GetChannelData<TypeT>(activity);
+                return true;
+            }
+            catch
```
willportnoy: >catch [](start = 12, length = 5)

please no "exceptions as control flow"

CSharp/Library/Microsoft.Bot.Connector.Shared/ActivityEx.cs
```diff
@@ -204,97 +296,57 @@ public bool HasContent()
         /// <summary>
         /// Get mentions 
         /// </summary>
+        /// <param name="activity"></param>
         /// <returns></returns>
-        public Mention[] GetMentions()
+        public static Mention[] GetMentions(this IMessageActivity activity)
```
willportnoy: >GetMentions [](start = 32, length = 11)

I like the move to extension methods

CSharp/Library/Microsoft.Bot.Connector.Shared/ActivityEx.cs
willportnoy: (sorry, just realized this was code motion)

---
In reply to: [113973797](https://github.com/Microsoft/BotBuilder/pull/2457#discussion_r113973797) [](ancestors = 113973797)

CSharp/Library/Microsoft.Bot.Connector.Shared/ConnectorAPI/Models/Activity.cs
```diff
@@ -133,13 +133,13 @@ public Activity(string type = default(string), string id = default(string), Date
         /// Array of address added
         /// </summary>
         [JsonProperty(PropertyName = "membersAdded")]
-        public IList<ChannelAccount> MembersAdded { get; set; }
+        public IList<ChannelAccount> MembersAdded { get; set; } = new ChannelAccount[0];
```
willportnoy: >new ChannelAccount[0] [](start = 66, length = 21)

Array.Empty<ChannelAccount>()

CSharp/Library/Microsoft.Bot.Connector.Shared/ConnectorAPI/Models/AudioCard.cs
```diff
@@ -73,13 +73,13 @@ public AudioCard(string title = default(string), string subtitle = default(strin
         /// Array of media Url objects
         /// </summary>
         [JsonProperty(PropertyName = "media")]
-        public IList<MediaUrl> Media { get; set; }
+        public IList<MediaUrl> Media { get; set; } = new MediaUrl[0];
```
willportnoy: >MediaUrl [](start = 57, length = 8)

while I generally like this change, it seems like it will only apply on outgoing activities, because the json deserializer won't do the same for incoming activities?

CSharp/Library/Microsoft.Bot.Connector.Shared/StateAPI/Models/BotData.cs
```diff
@@ -21,9 +21,8 @@ public partial class BotData
         /// <summary>
         /// Initializes a new instance of the BotData class.
         /// </summary>
-        public BotData(string eTag = default(string), object data = default(object))
+        public BotData(string eTag = @"*", object data = default(object))
```
willportnoy: >@"*" [](start = 37, length = 4)

default is "overwrite existing data because this node isn't aware of concurrent change made by other nodes" ?

willportnoy:
:clock1:

CSharp/Library/Microsoft.Bot.Builder/Dialogs/PromptDialog.cs
brandonh-msft: Without this you end up with an Ambiguous Reference exception.

CSharp/Library/Microsoft.Bot.Builder/ConnectorEx/BotData.cs
brandonh-msft: This change was for consistency. Many other places in the SDK we have this pattern; this unifies it in places partners were questioning the difference.

CSharp/Library/Microsoft.Bot.Builder/ConnectorEx/BotToUser.cs
brandonh-msft: probably a merge error.

CSharp/Library/Microsoft.Bot.Connector.Shared/ActivityEx.cs
brandonh-msft: What would be the appropriate way to achieve this behavior without duplicating the code w/in `GetChannelData`?

CSharp/Library/Microsoft.Bot.Connector.Shared/ConnectorAPI/Models/AudioCard.cs
brandonh-msft: which is the intent. The problem is when devs do `new()` and `.Add` on the `Media` property, it blows up because `.Media` is `null`

CSharp/Library/Microsoft.Bot.Connector.Shared/StateAPI/Models/BotData.cs
brandonh-msft: apparently. @matvelloso and I were told this should be what devs do when playing with the UserData via the State Client in a discussion regarding this behavior during a hackfest a number of weeks ago. "always set etag to '*'"

CSharp/Library/Microsoft.Bot.Connector.Shared/ActionTypes.cs
brandonh-msft: from https://msdn.microsoft.com/en-us/microsoft-teams/botsmessages#action---invoke-new] as highlighted in a [confused StackOverflow user's post](http://stackoverflow.com/questions/43623385/cardaction-button-with-type-invoke-doesnt-work-in-microsoft-bot-framework)

image

CSharp/Library/Microsoft.Bot.Connector.Shared/ActivityEx.cs
brandonh-msft: `IMessageActivity` requires `HasContent()` and `GetMentions()` - didn't want to go changing an interface.

CSharp/Library/Microsoft.Bot.Connector.Shared/ConnectorAPI/Models/Activity.cs
brandonh-msft: not available in .Net Core. Would `Enumerable.Empty<T>().ToList()` be acceptable? seems like too much work.

brandonh-msft:
🕐

CSharp/Library/Microsoft.Bot.Builder/ConnectorEx/BotData.cs
msft-shahins: Better change might be to removing default arguments from other interfaces rather than adding default(CancellationToken) to all interfaces. 

---
In reply to: [113984240](https://github.com/Microsoft/BotBuilder/pull/2457#discussion_r113984240) [](ancestors = 113984240)

CSharp/Library/Microsoft.Bot.Builder/ConnectorEx/BotData.cs
```diff
@@ -144,7 +142,7 @@ private static void ValidateETag(BotData botData, string value)
             }
         }
 
-        Task<bool> IBotDataStore<BotData>.FlushAsync(IAddress key, CancellationToken cancellationToken)
+        Task<bool> IBotDataStore<BotData>.FlushAsync(IAddress key, CancellationToken cancellationToken )
```
msft-shahins: >  [](start = 102, length = 1)

remove extra space added after the last argument.

CSharp/Library/Microsoft.Bot.Builder/ConnectorEx/BotToUser.cs
```diff
@@ -31,14 +31,14 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 //
 
+using Microsoft.Bot.Builder.ConnectorEx;
 using System;
```
msft-shahins: sort usings.

CSharp/Library/Microsoft.Bot.Builder/Dialogs/IDialogContext.cs
```diff
@@ -1,91 +1,91 @@
...
-
-    /// <summary>
-    /// The context for the execution of a dialog's conversational process.
-    /// </summary>
-    public interface IDialogContext : IDialogStack, IBotContext
-    {
-    }
-
+using System;
+using Microsoft.Bot.Builder.Dialogs.Internals;
```
msft-shahins: Sort usings and we recently started using the following VS rule when sort usings: Place 'System' directives first when sorting usings

msft-shahins:
> namespace Microsoft.Bot.Connector

All of the files in ConnectorAPI/Models are autorest generated files. If you want to make changes to it, it is better to change the swagger file. 

---
Refers to: CSharp/Library/Microsoft.Bot.Connector.Shared/ConnectorAPI/Models/Activity.cs:5 in 49f8eae. [](commit_id = 49f8eaeadeeeffdf704812838ee9953cf909b909, deletion_comment = False)

CSharp/Library/Microsoft.Bot.Builder/Resource/Resources.fr.resx
```diff
@@ -1,251 +1,251 @@
...
-    <value>un, une, la</value>
-  </data>
-  <data name="LanguageStopWords" xml:space="preserve">
-    <value></value>
-  </data>
-  <data name="MatchCurrentChoice" xml:space="preserve">
-    <value>Le choix actuel ("c"); courant; c; choix actuel</value>
-  </data>
-  <data name="MatchNo" xml:space="preserve">
-    <value>Non;pas;rien;2</value>
```
msft-shahins: >Non [](start = 11, length = 3)

these files are generated from xlf files. Make changes in xlf files and use Microsoft multilingual app toolkit to generate resources files.

CSharp/Library/Microsoft.Bot.Connector.NetFramework/BotAuthentication.cs
```diff
@@ -60,7 +55,7 @@ public class BotAuthentication : ActionFilterAttribute
 
 
 
-        public override async Task OnActionExecutingAsync(HttpActionContext actionContext, CancellationToken cancellationToken)
+        public override async Task OnActionExecutingAsync(HttpActionContext actionContext, CancellationToken cancellationToken = default(CancellationToken))
```
msft-shahins: >default [](start = 129, length = 7)

Remove default(CancellationToken)

CSharp/Library/Microsoft.Bot.Builder/Dialogs/PromptDialog.cs
```diff
@@ -172,24 +172,24 @@ protected string DefaultTooManyAttempts
...
         /// <param name="attempts"> Maximum number of attempts.</param>
         /// <param name="promptStyler"> The prompt styler.</param>
-        /// <param name="descriptions">Descriptions for each prompt.</param>
+        /// <param name="titles">Descriptions for each prompt.</param>
         /// <param name="speak"> The Speak tag (SSML markup for text to speech).</param>
         /// <param name="retrySpeak"> What to display on retry Speak (SSML markup for text to speech).</param>
         /// <param name="recognizer"> Entity Recognizer to parse the message content.</param>
-        public PromptOptions(string prompt, string retry = null, string tooManyAttempts = null, IReadOnlyList<T> options = null, int attempts = 3, PromptStyler promptStyler = null, IReadOnlyList<string> descriptions = null, string speak = null, string retrySpeak = null, IPromptRecognizer recognizer = null)
+        public PromptOptions(string prompt, IReadOnlyList<string> titles = null, IReadOnlyList<T> values = null, string retry = null, string tooManyAttempts = null, int attempts = 3, PromptStyler promptStyler = null, string speak = null, string retrySpeak = null, IPromptRecognizer recognizer = null)
```
msft-shahins: This is a breaking change and probably better not to reorder the parameters in the constructor.

CSharp/Library/Microsoft.Bot.Builder/Dialogs/PromptDialog.cs
```diff
@@ -199,24 +199,24 @@ public PromptOptions(string prompt, string retry = null, string tooManyAttempts
...
         /// <param name="promptStyler"> The prompt styler.</param>
-        /// <param name="descriptions">Descriptions for each prompt.</param>
+        /// <param name="titles">Descriptions for each prompt.</param>
         /// <param name="speak"> The Speak tag (SSML markup for text to speech).</param>
         /// <param name="retrySpeak"> What to display on retry Speak (SSML markup for text to speech).</param>
         /// <param name="recognizer"> Entity Recognizer to parse the message content.</param>
-        public PromptOptions(string prompt, string retry = null, string tooManyAttempts = null, IReadOnlyDictionary<T, IReadOnlyList<T>> choices = null, int attempts = 3, PromptStyler promptStyler = null, IReadOnlyList<string> descriptions = null, string speak = null, string retrySpeak = null, IPromptRecognizer recognizer = null)
+        public PromptOptions(string prompt, IReadOnlyList<string> titles = null, IReadOnlyDictionary<T, IReadOnlyList<T>> values = null, string retry = null, string tooManyAttempts = null, int attempts = 3, PromptStyler promptStyler = null, string speak = null, string retrySpeak = null, IPromptRecognizer recognizer = null)
```
msft-shahins: >IReadOnlyList [](start = 44, length = 13)

same here.

msft-shahins:
>     public partial class BotData

All the file under StateAPI/Models are autorest generated files and better not to change them directly 

---
Refers to: CSharp/Library/Microsoft.Bot.Connector.Shared/StateAPI/Models/BotData.cs:14 in 49f8eae. [](commit_id = 49f8eaeadeeeffdf704812838ee9953cf909b909, deletion_comment = False)

CSharp/Library/Microsoft.Bot.Connector.Shared/StateAPI/Models/BotData.cs
```diff
@@ -21,9 +21,8 @@ public partial class BotData
...
         {
-            ETag = eTag;
```
msft-shahins: >ETag [](start = 12, length = 4)

why removing the setter?

CSharp/Samples/PizzaBot/PizzaOrderDialog.cs
```diff
@@ -108,7 +108,7 @@ enum Days { Saturday, Sunday, Monday, Tuesday, Wednesday, Thursday, Friday };
 
             var days = (IEnumerable<Days>)Enum.GetValues(typeof(Days));
             PromptDialog.Choice(context, StoreHoursResult, days, "Which day of the week?",
-                descriptions: from day in days
```
msft-shahins: >descriptions [](start = 16, length = 12)

Introduction of breaking changes to primitive dialogs of SDK, e.g. PromptDialogs, should be minimized!

msft-shahins:
:clock1:

CSharp/Library/Microsoft.Bot.Connector.Shared/ActionTypes.cs
msft-shahins: invoke is not an action type. @dandriscoll can you confirm?

brandonh-msft:
i'm going to break this thing up.

Issue #1885, opened 2016-12-19
NodeJS Calling SDK should allow for clock skew when verifying auth token


ctstone:
BotBuilder version is 3.0.1 for Node

Today I started seeing my call connector return HTTP 403 responses to Skype. I tracked the responses to `CallConnector.prototype.verifyBotFramework` which is checking the auth token NBF field against current time with no clock skew. NBF is checked in at least 2 places: once explicitly in code and another via a call to `jsonwebtoken.verify` (to my eyes, the first check is redundant) 

The jsonwebtoken library supports clock skew as an option, so it would be great to see this exposed as an option on `CallConnector`. Or simply choose a reasonable default skew allowance, per the spec:

> The nbf (not before) claim identifies the time before which the token MUST NOT be accepted for processing. The processing of the nbf claim requires that the current date/time MUST be after or equal to the not-before date/time listed in the nbf claim. **Implementers MAY provide for some small leeway, usually no more than a few minutes, to account for clock skew**.

dandriscoll:
Agree, this should be relying on the clock skew option in jsonwebtoken.

ctstone:
See new PR at #2502 for a more complete fix

PR #2502, opened 2017-03-27
[Node SDK] Fix for clock skew in #1885

ctstone:
This fix extends the prior clock skew fix. 

EXP and NBF are checked manually before passing over to the JWT lib, but the prior fix only applied to the later JWT validation call. This fix ensures that current time is adjusted for the manual check.

A few notes

1. It seems odd to keep the JS files in the source. Is that intentional? I updated CallConnector.js from the generated TS and it looks like the JS was already out of date before my changes.
1. I moved the clock tolerance value to a constant since it is used in multiple places
1. Why manually validate the token if you are going to pass it on the JWT lib? Can probably simplify this to only validate once.


Node/calling/src/bots/CallConnector.ts
```diff
@@ -198,7 +199,7 @@ export class CallConnector implements ucb.ICallConnector, bs.IBotStorage {
 
                     // verify appId, issuer, token expirs and token notBefore
                     if (decoded.payload.aud != this.settings.appId || decoded.payload.iss != issuer || 
-                        now > decoded.payload.exp || now < decoded.payload.nbf) {
+                        (now - CLOCK_TOLERANCE) > decoded.payload.exp || (now + CLOCK_TOLERANCE) < decoded.payload.nbf) {
```
msft-shahins: you can remove  `(now - CLOCK_TOLERANCE) > decoded.payload.exp || (now + CLOCK_TOLERANCE) < decoded.payload.nbf)` check from here and rely on the clockTolerance check done by jsonwebtoken module.

ctstone:
Closing this PR. I will push a cleaner fix on its own branch

Issue #1530, opened 2016-10-26
[Known Issue] Why does BotBuilder.Prompts.number method parse like this?


jorge-sanz:
I am using `BotBuilder.Prompts.number` method. 

What is the purpose of parsing like this (check bellow) and not reject it and consider it like something that is not a real number? 
Input: `v66v6` -> Parsed answer: `66`

msft-shahins:
based on the syntax I assume that this is a node.js bug. Can you confirm this?

Stevenic:
The RegEx we're using doesn't look to be strict enough. I'll try to improve it.

chrimc62:
@Stevenic or @maxpert did 3.5 fix this?

Stevenic:
It's not in 3.5. I'll have to work on it for 3.6

nwhitmont:
[Bugfix] PR: https://github.com/Microsoft/BotBuilder/pull/2553

stevengum97:
Enhancements for number parser still needed, reopening issue.

nwhitmont:
@Stevenic Any updates on the number recognizer fix?

nwhitmont:
relates to Intercom#15644

nwhitmont:
Currently, this is expected behavior.

PR #2553, opened 2017-04-05
[Node SDK] #1530 update EntityRecognizer.parseNumer() numberExp regex

nwhitmont:
Fix for #1530 - Update EntityRecognizer.parseNumber() regex contained in numberExp to `/^[0-9]*$/`


Stevenic:
This is too restrictive... It doesn't support decimal numbers and we want to be able to recognize numbers in utterances like "I'd like 5 please".  We may need to address this issue using more than just RegEx's.

Issue #1810, opened 2016-12-07
PromptDialog.PromptConfirm localization issue


Taras-Tyrsa:
BotBuilder v.3.3.3.0

Microsoft.Bot.Builder.Dialogs.PromptDialog.PromptConfirm class has static initialization for Options and Patterns, which makes impossible to change language once initialized:

```
public sealed class PromptConfirm : Prompt<bool, string>
{
...
    /// <summary>
    /// The yes, no options for confirmation prompt
    /// </summary>
    public static string[] Options { set; get; } = { Resources.MatchYes.SplitList().First(), Resources.MatchNo.SplitList().First() };

    /// <summary>
    /// The patterns for matching yes/no responses in the confirmation prompt.
    /// </summary>
    public static string[][] Patterns { get; set; } = { Resources.MatchYes.SplitList(), Resources.MatchNo.SplitList() };
```

It will take current culture during first usage and won't change it afterwards:
```
Thread.CurrentThread.CurrentCulture = CultureInfo.GetCultureInfo("de-DE");
Thread.CurrentThread.CurrentUICulture = CultureInfo.GetCultureInfo("de-DE");

var options = PromptDialog.PromptConfirm.Options;
// options = [ 'Ja', 'Nein' ]

Thread.CurrentThread.CurrentCulture = CultureInfo.GetCultureInfo("en-US");
Thread.CurrentThread.CurrentUICulture = CultureInfo.GetCultureInfo("en-US");

options = PromptDialog.PromptConfirm.Options;
// EXPECTED: options = [ 'Yes', 'No' ]
// ACTUAL: options = [ 'Ja', 'Nein' ]
```

Taras-Tyrsa:
Just a little remark: manual set of Options and Patterns can lead to race conditions when two parallel requests use it (e.g. one request sets English translations and another one - German).

Stevenic:
@hailiu2586 This is an issue with C# version of SDK.

willportnoy:
@Taras-Tyrsa those are static property getters (not static fields), so that code executes every time the "method" is called.  Are they being cached elsewhere?

Taras-Tyrsa:
`public static string[] Options { set; get; } = { Resources.MatchYes.SplitList().First(), Resources.MatchNo.SplitList().First() };`

is not the same as

```
public static string[] Options {
    get { return new [] { Resources.MatchYes.SplitList().First(), Resources.MatchNo.SplitList().First() }; }
};
```

[...]

PR #2564, opened 2017-04-06
Update PromptConfirm's Options and Patterns getter

ejadib:
This address the issue #1810. Now the code will be executed every time the property is called.

### DIFF
--- a/CSharp/Library/Microsoft.Bot.Builder/Dialogs/PromptDialog.cs
+++ b/CSharp/Library/Microsoft.Bot.Builder/Dialogs/PromptDialog.cs
@@ -425,6 +425,9 @@ namespace Microsoft.Bot.Builder.Dialogs
         [Serializable]
         public sealed class PromptConfirm : Prompt<bool, string>
         {
+            private static string[] options;
+            private static string[][] patterns;
+
             /// <summary>
             /// Index of yes descriptions.
             /// </summary>
@@ -442,7 +445,15 @@ namespace Microsoft.Bot.Builder.Dialogs
             {
                 get
                 {
-                    return new string[] { Resources.MatchYes.SplitList().First(), Resources.MatchNo.SplitList().First() };
+                    if (options == null)
+                    {
+                        return new string[] { Resources.MatchYes.SplitList().First(), Resources.MatchNo.SplitList().First() };
+                    }
+                    return options;
+                }
+                set
+                {
+                    options = value;
                 }
             }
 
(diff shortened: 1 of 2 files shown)


CSharp/Library/Microsoft.Bot.Builder/Dialogs/PromptDialog.cs
```diff
@@ -434,16 +434,28 @@ public sealed class PromptConfirm : Prompt<bool, string>
             /// Index of no descriptions.
             /// </summary>
             public const int No = 1;
-
+            
             /// <summary>
             /// The yes, no options for confirmation prompt
             /// </summary>
-            public static string[] Options { set; get; } = { Resources.MatchYes.SplitList().First(), Resources.MatchNo.SplitList().First() };
```
msft-shahins: removing the setter can be a breaking change. Is it possible to fix the bug without removing the setter?

CSharp/Library/Microsoft.Bot.Builder/Dialogs/PromptDialog.cs
```diff
@@ -434,16 +434,28 @@ public sealed class PromptConfirm : Prompt<bool, string>
...
+                get
+                {
+                    return new string[] { Resources.MatchYes.SplitList().First(), Resources.MatchNo.SplitList().First() };
+                }
+            }
 
             /// <summary>
             /// The patterns for matching yes/no responses in the confirmation prompt.
             /// </summary>
-            public static string[][] Patterns { get; set; } = { Resources.MatchYes.SplitList(), Resources.MatchNo.SplitList() };
```
msft-shahins: same as above.

CSharp/Library/Microsoft.Bot.Builder/Dialogs/PromptDialog.cs
ejadib: We could leave the setter by having the a private variable that hold the value set and then in the getter have some logic that returns the variable or the defaults

```
public static string[] Options
{
    get
    {
        if (this.options == null)
        {
            return new string[] { Resources.MatchYes.SplitList().First(), Resources.MatchNo.SplitList().First() };
        }

        return this.options;
    }
    set
    {
        this.options = value;
    }
}
```

Though it seems better to just make these properties private and non-static properties and have parameters in the constructor that overrides them when provided. If they are not provided we will default to the values in the Resources. Though this will imply some change in the static Confirm method of the PromptDialog class since it's using th[e static Options property](https://github.com/southworkscom/BotBuilder/blob/master/CSharp/Library/Microsoft.Bot.Builder/Dialogs/PromptDialog.cs#L298).

ejadib: @msft-shahins We just pushed the updates to address the feedback

PR #2569, opened 2017-04-07
[Node SDK] feat: support thumbsup/down as yes/no response

### LINKED ISSUE #2069: Support thumbsup/thumbsdown emojis
The [current EntityRecognizer](https://github.com/Microsoft/BotBuilder/blob/master/Node/core/src/dialogs/EntityRecognizer.ts#L65-L66) does not support thumbsup and thumbsdown emojis (👍 and 👎). It would be interesting to add them to the valid yes/no answers.

If you think it's a good idea I can send a PR.

palmerabollo:
Resolves #2069


nwhitmont:
duplicate, per @Stevenic

Issue #2590, opened 2017-04-11
Instantiating ConnectorClient using connector 3.6.0-alpha fails


msft-shahins:
`new ConnectorClient(serviceUrl)` fails with the following exception:``The service provider instance was not register. Please call RegisterServiceProvider before using ServiceProvider.Instance.``

This happens when Microsoft.Bot.Connector is instantiated directly and BotAuthentication is not referenced on the code path. An example, will be a worker job that runs independently and gets triggered to instantiate connector client to send proactive messages to user. Because of not referencing BotServiceProvider or BotAuthentication on the code (and builder container is not referenced), instantiation of ConnectorClient will fail since no ServiceProvider instance is initialized: https://github.com/Microsoft/BotBuilder/blob/develop/CSharp/Library/Microsoft.Bot.Connector.Shared/ServiceProvider.cs#L82

Also there is a typo in the InvalidOperationException message: 
https://github.com/Microsoft/BotBuilder/blob/develop/CSharp/Library/Microsoft.Bot.Connector.Shared/ServiceProvider.cs#L86

luizcarlosfaria:
Just call `services.UseBotConnector()` on ConfigureServices method of your Startup class.
This method is provided by Microsoft.Bot.Connector.BotServiceProviderExtensions extension class.

PR #2591, opened 2017-04-12
[C# SDK] Improve logic to auto register ASPNET service provider

andreesteve:
Description of problem
----------
#2590 

Usage of ASPNET BotConnector without built-in authentication or BotBuilder would require manually registering *Microsoft.Bot.Connector.BotServiceProvider* with the common library, which breaks backward compatibility.

Proposed solution
----------
Currently auto-registration of ASPNET connector happens in two places [DialogModule.cs](https://github.com/Microsoft/BotBuilder/blob/develop/CSharp/Library/Microsoft.Bot.Builder.Autofac/Dialogs/DialogModule.cs#L423) and from static constructor in [BotServiceProvider](https://github.com/Microsoft/BotBuilder/blob/develop/CSharp/Library/Microsoft.Bot.Connector.NetFramework/BotServiceProvider.cs#L39), the latter being leveraged by BotAuthentication. This implies that only usages with authentication and/or BotBuilder would be backward compatible (i.e. not require manual registration of the ASPNET connector service provider).

To circumvent that limitation, the auto-registration was moved to the common portable library and it is done through reflection.

With this change, AspNet Core service provider is still required to be registered on Startup, as before.

Testing
----------

- [x] Unit tests passing (exception of PromptSuccess_Confirm_Yes and PromptSuccess_Confirm_Yes_CaseInsensitive that fail with or without these changes)
- [x] AspNet with bot builder with authentication
- [x] AspNet with bot builder without authentication
- [x] AspNet without bot builder with authentication
- [x] AspNet without bot builder without authentication
- [x] AspNetCore samples

### DIFF
--- a/CSharp/Library/Microsoft.Bot.Builder.Autofac/Dialogs/DialogModule.cs
+++ b/CSharp/Library/Microsoft.Bot.Builder.Autofac/Dialogs/DialogModule.cs
@@ -410,7 +410,7 @@ namespace Microsoft.Bot.Builder.Dialogs.Internals
         /// <param name="builder">The builder used to register dependencies.</param>
         private void RegisterBotConnectorImplementation(ContainerBuilder builder)
         {
-            // service provider is assumed to be availble at this point
+            // service provider is assumed to be available at this point
             // For ASPNET platform, it will be auto-registered
             // For all other platforms, consumer is expected to set up provider during startup
             builder.Register<ServiceProvider>(c => ServiceProvider.Instance)
--- a/CSharp/Library/Microsoft.Bot.Builder.Autofac/Dialogs/DialogModule.cs
+++ b/CSharp/Library/Microsoft.Bot.Builder.Autofac/Dialogs/DialogModule.cs
@@ -410,35 +410,12 @@ namespace Microsoft.Bot.Builder.Dialogs.Internals
         /// <param name="builder">The builder used to register dependencies.</param>
         private void RegisterBotConnectorImplementation(ContainerBuilder builder)
         {
-            if (ServiceProvider.IsRegistered)
-            {
-                builder.Register<ServiceProvider>(c => ServiceProvider.Instance)
-                    .AsSelf()
-                    .SingleInstance();
-            }
-            else
-            {
-                // Finds the bot connector being used in this application domain
-                // that is not the common connector assembly
-                System.Reflection.Assembly connectorAssembly = AppDomain.CurrentDomain.
-                    GetAssemblies()
-                    .Where(a => a.FullName.StartsWith("Microsoft.Bot.Connector"))
-                    .Where(a => a != typeof(BotData).Assembly)
-                    .FirstOrDefault();
-
-                if (connectorAssembly != null)
-                {
-                    // register the IServiceProvider instance provided in the connector with the Common connector
-                    builder.RegisterAssemblyTypes(connectorAssembly)
-                        .Where(t => t.IsAssignableTo<IServiceProvider>())
-                        .AsImplementedInterfaces()
-                        .SingleInstance();
-
-                    builder.Register<ServiceProvider>(c => { if (!ServiceProvider.IsRegistered) { ServiceProvider.RegisterServiceProvider(c.Resolve<IServiceProvider>()); } return ServiceProvider.Instance; })
-                        .AsSelf()
-                        .SingleInstance();
-                }
-            }
+            // service provider is assumed to be availble at this point
+            // For ASPNET platform, it will be auto-registered
+            // For all other platforms, consumer is expected to set up provider during startup
+            builder.Register<ServiceProvider>(c => ServiceProvider.Instance)
+                .AsSelf()
+                .SingleInstance();
         }
     }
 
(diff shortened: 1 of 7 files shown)


CSharp/Library/Microsoft.Bot.Connector.Shared/ServiceProvider.cs
```diff
@@ -99,5 +107,49 @@ private static void ThrowOnNullInstance()
...
+            Assembly connectorAssembly = null;
+
+            try
+            {
+                var connectorAssemblyName = new AssemblyName(AspNetBotConnectorAssemblyName);
+                connectorAssembly = Assembly.Load(connectorAssemblyName);
+            }
+            catch (Exception)
+            {
+                // assembly not available
```
iedeny: maybe we should at least Debug.Writeline() here..

CSharp/Samples/SimpleEchoBot/App_Start/WebApiConfig.cs
```diff
@@ -32,6 +33,12 @@ public static void Register(HttpConfiguration config)
                 routeTemplate: "api/{controller}/{id}",
                 defaults: new { id = RouteParameter.Optional }
             );
+
+            config.Routes.MapHttpRoute(
+                name: "ConnectorOnlytApi",
```
iedeny: "ConnectorOnlyApi"? (remove 't')

CSharp/Library/Microsoft.Bot.Connector.Shared/ServiceProvider.cs
```diff
@@ -99,5 +107,49 @@ private static void ThrowOnNullInstance()
...
+                // cannot log, because we don't have service provider to get logger from
+            }
+
+            if (connectorAssembly != null)
+            {
+                Type botServiceProviderType = connectorAssembly.DefinedTypes
+                        .FirstOrDefault(t => t.FullName.Equals(AspNetBotConnectorServiceProviderFullQualifiedName))
```
willportnoy: >t.FullName.Equals(AspNetBotConnectorServiceProviderFullQualifiedName) [](start = 45, length = 69)

use [https://msdn.microsoft.com/en-us/library/system.reflection.assemblyname.name(v=vs.110).aspx](https://msdn.microsoft.com/en-us/library/system.reflection.assemblyname.name(v=vs.110).aspx)

CSharp/Library/Microsoft.Bot.Builder.Autofac/Dialogs/DialogModule.cs
```diff
@@ -410,35 +410,12 @@ protected override void Load(ContainerBuilder builder)
...
-                        .Where(t => t.IsAssignableTo<IServiceProvider>())
-                        .AsImplementedInterfaces()
-                        .SingleInstance();
-
-                    builder.Register<ServiceProvider>(c => { if (!ServiceProvider.IsRegistered) { ServiceProvider.RegisterServiceProvider(c.Resolve<IServiceProvider>()); } return ServiceProvider.Instance; })
-                        .AsSelf()
-                        .SingleInstance();
-                }
-            }
+            // service provider is assumed to be availble at this point
```
willportnoy: >availble [](start = 49, length = 8)

spelling

CSharp/Library/Microsoft.Bot.Connector.Shared/ServiceProvider.cs
```diff
@@ -99,5 +107,49 @@ private static void ThrowOnNullInstance()
...
+        private static void TryAutoRegisterForBackwardCompatibility()
+        {
+            const string AspNetBotConnectorAssemblyName = "Microsoft.Bot.Connector";
+            const string AspNetBotConnectorServiceProviderFullQualifiedName = "Microsoft.Bot.Connector.BotServiceProvider";
+
+            Assembly connectorAssembly = null;
+
+            try
+            {
+                var connectorAssemblyName = new AssemblyName(AspNetBotConnectorAssemblyName);
```
willportnoy: >AspNetBotConnectorAssemblyName [](start = 61, length = 30)

will this work with strong names?

andreesteve: Yes. I ended up testing it to make sure. Assembly.Load doesn't require the token key when loading signed binaries. We could add it, though, to prevent loading a unsigned version of the assembly, but I'm afraid it would prevent any one that doesn't have the key to work unsigned builds.

CSharp/Library/Microsoft.Bot.Connector.Shared/ServiceProvider.cs
andreesteve: The "t.FullName" here is for the type full name (i.e. namespace + class name), not the assembly's.

CSharp/Library/Microsoft.Bot.Builder.Autofac/Dialogs/DialogModule.cs
andreesteve: Thanks! Fixed.

CSharp/Samples/SimpleEchoBot/App_Start/WebApiConfig.cs
andreesteve: Thanks. Fixed.

EricDahlvang:

PR #2596, opened 2017-04-12
[C# SDK] Add SayAsync extension method

ejadib:
Some things to provide feedback on:

- Location of the MessageOptions class
- Location of the unit tests for the SayAsync extension method

### DIFF
--- a/CSharp/Library/Microsoft.Bot.Builder/Dialogs/IDialogContext.cs
+++ b/CSharp/Library/Microsoft.Bot.Builder/Dialogs/IDialogContext.cs
@@ -161,7 +161,7 @@ namespace Microsoft.Bot.Builder.Dialogs
         /// <param name="botToUser">Communication channel to use.</param>
         /// <param name="text">The message text.</param>
         /// <param name="speak">The SSML markup for text to speech.</param>
-        /// <param name="options">The optioms for the message.</param>
+        /// <param name="options">The options for the message.</param>
         /// <param name="locale">The locale of the text.</param>
         /// <param name="cancellationToken">The cancellation token.</param>
         /// <returns>A task that represents the post operation.</returns>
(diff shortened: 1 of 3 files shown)


CSharp/Library/Microsoft.Bot.Builder/Dialogs/IDialogContext.cs
```diff
@@ -109,6 +149,44 @@ public static partial class Extensions
...
+        /// <summary>
+        /// Post a message and optional SSML to be sent to the user, using previous messages to establish a conversation context.
+        /// </summary>
+        /// <remarks>
+        /// If the locale parameter is not set, locale of the incoming message will be used for reply.
+        /// </remarks>
+        /// <param name="botToUser">Communication channel to use.</param>
+        /// <param name="text">The message text.</param>
+        /// <param name="speak">The SSML markup for text to speech.</param>
+        /// <param name="options">The optioms for the message.</param>
```
willportnoy: "options" spelling

CSharp/Library/Microsoft.Bot.Builder/Dialogs/IDialogContext.cs
```diff
@@ -109,6 +149,44 @@ public static partial class Extensions
...
+
+            message.Text = text;
+            message.Speak = speak;
+
+            if (!string.IsNullOrEmpty(locale))
+            {
+                message.Locale = locale;
+            }
+
+            if (options != null)
```
willportnoy: is it it better to package these into a class or to use keyword named arguments?

CSharp/Library/Microsoft.Bot.Builder/Dialogs/IDialogContext.cs
ejadib: fixed

CSharp/Library/Microsoft.Bot.Builder/Dialogs/IDialogContext.cs
ejadib: Are you referring to the MessageOptions class? I’m fine either way, I added a class so it’s aligned to what Steve did. He has an interface [IMessageOptions](https://github.com/Microsoft/BotBuilder/blob/master/Node/core/lib/botbuilder.d.ts#L106-L122). We can change it to keyword named arguments if that makes more sense. LMK

PR #2623, opened 2017-04-19
Fix index when dealing with reverse ordinals

ejadib:
When dealing with reverse ordinals, the `match.entity` will be a negative value and so it needs to be subtracted from the `choices.length` to get the proper index. The current code is adding the `match.entity` value because its doing minus negative number.


ejadib:
@Stevenic FYI

Stevenic:
cood catch

PR #2672, opened 2017-05-01
Api extensions

brandonh-msft:
Adding some extensions to the API by way of extension methods and additions to the current API that make it more cohesive.


willportnoy:
:shipit:

PR #2673, opened 2017-05-01
No null collections

brandonh-msft:
Collections off our activity and card objects should be able to have `.Add()` run on them readily; today they default to `null` values which causes a first-time-crash almost without fail for developers getting to learn our API.


willportnoy:
:shipit:

PR #2674, opened 2017-05-01
[C# SDK] Add delegate to build method to intercept messages before send in for…

rcarubbi:
I need to intercept answers while formflow are filling an entity to use a voice recognition service in order to treat each answer. 

For each answer sent by user I'll get the attachment, decode to PCM and send to bing speech api once this is the format expected by it.

However, the formflow feature doesn't allow to intercept the answer sent by user before set the value in property.

With this solution, you can write some code in a delegate that will be invoked in this moment. #


chrimc62:
This functionality overlaps with the existing FormBuilder.Prompter PromptAsyncDelegate which lets you do exactly the same thing.

PR #2676, opened 2017-05-01
[C# SDK] ETag should default to *, not empty

brandonh-msft:
@matvelloso and I were told this should be what devs do when playing with the UserData via the State Client in a discussion regarding this behavior during a hackfest a number of weeks ago. "always set etag to '*'." This is also what is documented [here](https://docs.botframework.com/en-us/csharp/builder/sdkreference/stateapi.html):
image

### DIFF
--- a/CSharp/Library/Microsoft.Bot.Connector.Shared/StateAPI/Models/BotData.cs
+++ b/CSharp/Library/Microsoft.Bot.Connector.Shared/StateAPI/Models/BotData.cs
@@ -21,7 +21,7 @@ namespace Microsoft.Bot.Connector
         /// <summary>
         /// Initializes a new instance of the BotData class.
         /// </summary>
-        public BotData(string eTag = default(string), object data = default(object))
+        public BotData(string eTag = @"*", object data = default(object))
         {
             ETag = eTag;
             Data = data;


CSharp/Library/Microsoft.Bot.Connector.Shared/StateAPI/Models/BotData.cs
```diff
@@ -21,9 +21,8 @@ public partial class BotData
         /// <summary>
         /// Initializes a new instance of the BotData class.
         /// </summary>
-        public BotData(string eTag = default(string), object data = default(object))
+        public BotData(string eTag = @"*", object data = default(object))
         {
-            ETag = eTag;
```
msft-shahins: why removing setter from the constructor?

CSharp/Library/Microsoft.Bot.Connector.Shared/StateAPI/Models/BotData.cs
```diff
@@ -21,9 +21,8 @@ public partial class BotData
         /// <summary>
         /// Initializes a new instance of the BotData class.
         /// </summary>
-        public BotData(string eTag = default(string), object data = default(object))
+        public BotData(string eTag = @"*", object data = default(object))
```
msft-shahins: not sure if we want to default to "*" for etag since it increases the risk of data getting overwritten by mistake.

brandonh-msft: per my comment on the PR, this was direction we were given, and what's documented for devs to do...

CSharp/Library/Microsoft.Bot.Connector.Shared/StateAPI/Models/BotData.cs
brandonh-msft: not sure why that happened. fixed. thanks.

CSharp/Library/Microsoft.Bot.Connector.Shared/StateAPI/Models/BotData.cs
willportnoy: @RobStand how can we file an issue to remove this Etag star from the documentation?  It is incorrect guidance because it can cause data loss.

brandonh-msft:
@willportnoy / @RobStand 
if this PR is indeed incorrect, and the Docs will be changed to reflect this, feel free to close, np.

msft-shahins:
closing this since defaulting etag to "*" increases the risk of data incosistency. @RobStand can you fix the documentation issue that @brandonh-msft  is mentioning here?

CSharp/Library/Microsoft.Bot.Connector.Shared/StateAPI/Models/BotData.cs
RobStand: @willportnoy I'll find it in the docs and fix. You can file bugs in the docs repo.

kbrandl: I don't believe any updates are needed to our docs.  The only time ETag=* is mentioned in our docs is to say that ETag=* in GET response (i.e., GetUserData, GetConversationData, or GetPrivateConversationData) indicates that no state data exists in the store for the user and/or conversation that is specified in a "Get..." method call. The docs do not mention anything about setting ETag=* in a SET request.

brandonh-msft: is the link above not in our docs?

kbrandl: @brandonh-msft -- sorry for the confusion. We're in the process of finalizing a new doc set that will be published next week before build -- the [article that you linked to](https://docs.botframework.com/en-us/csharp/builder/sdkreference/stateapi.html) will be replaced by a new article at that time.  I've verified that the new article (about managing state data) does not specify that ETag should be set to asterisk (*) when saving state data.  So, this documentation issue will no longer exist once the new docs are published next week (and the old docs are decommissioned).

PR #2678, opened 2017-05-01
[C# SDK] context.EndConversation extension method

brandonh-msft:
for parity with node SDK

### DIFF
--- a/CSharp/Library/Microsoft.Bot.Builder/Dialogs/IDialogContext.cs
+++ b/CSharp/Library/Microsoft.Bot.Builder/Dialogs/IDialogContext.cs
@@ -150,6 +150,21 @@ namespace Microsoft.Bot.Builder.Dialogs
             await botToUser.PostAsync(message, cancellationToken);
         }
 
+        /// <summary>
+        /// Ends the conversation.
+        /// </summary>
+        /// <param name="botToUser">Communication channel to use.</param>
+        /// <param name="code">The conversation end code.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        /// <returns></returns>
+        public static async Task EndConversation(this IBotToUser botToUser, string code = EndOfConversationCodes.CompletedSuccessfully, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            var message = botToUser.MakeMessage();
+            message.Type = ActivityTypes.EndOfConversation;
+            message.AsEndOfConversationActivity().Code = code;
+
+            await botToUser.PostAsync(message, cancellationToken);
+        }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


CSharp/Library/Microsoft.Bot.Builder/Dialogs/IDialogContext.cs
```diff
@@ -151,6 +151,21 @@ public static partial class Extensions
...
         }
 
+        /// <summary>
+        /// Ends the conversation.
+        /// </summary>
+        /// <param name="botToUser">Communication channel to use.</param>
+        /// <param name="code">The conversation end code.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        /// <returns></returns>
+        public static async Task EndConversation(this IBotToUser botToUser, string code = EndOfConversationCodes.CompletedSuccessfully, CancellationToken cancellationToken = default(CancellationToken))
```
msft-shahins: EndConversation is a misleading name. Probably PostEndConversation or SayEndCoversation is a better name for this method, since it is not impacting the stack and only sends an EndOfConversation Activity to channel.

CSharp/Library/Microsoft.Bot.Builder/Dialogs/IDialogContext.cs
```diff
@@ -151,6 +150,36 @@ public static partial class Extensions
...
+
+            // Send EndOfConversation message
+            var eocMessage = context.MakeMessage();
+            eocMessage.Type = ActivityTypes.EndOfConversation;
+            eocMessage.AsEndOfConversationActivity().Code = code;
+
+            await context.PostAsync(eocMessage, cancellationToken);
+
+            // Clear the stack
+            context.Reset();
```
willportnoy: It's not appropriate for a called child dialog to "tear down the entire stack" - it would be more appropriate to throw an exception and let it propagate up the stack, and then the root looping dialog could clear out conversation data.

(separately, I'm not sure it works to clear out the dialog stack from within a dialog)

willportnoy:
:clock1:

CSharp/Library/Microsoft.Bot.Builder/Dialogs/IDialogContext.cs
brandonh-msft: hm. this code does (as far as i can tell) the same thing the Node SDK is doing so.... how is it possible/appropriate there but not here?

@Stevenic - any insight on the Node side?

brandonh-msft:
Well, I'm happy to have this be the starting point for a *real* implementation of this method since our C# developers are asking for it. Let me know if there's anything I can do!

PR #2687, opened 2017-05-02
[C# SDK] Add IEquatable to Attachment and Do not respond to Ping in SimpleDispatchDialog sample

EricDahlvang:
Other types on Activity already implement this interface.

### DIFF
--- a/CSharp/Library/Microsoft.Bot.Connector.Shared/AttachmentEx.cs
+++ b/CSharp/Library/Microsoft.Bot.Connector.Shared/AttachmentEx.cs
@@ -1,0 +1,34 @@
+﻿namespace Microsoft.Bot.Connector
+{
+    using System;
+
+    public partial class Attachment : IEquatable<Attachment>
+    {
+        public bool Equals(Attachment other)
+        {
+            return other != null
+                && this.ContentType == other.ContentType
+                && this.ContentUrl == other.ContentUrl
+                && object.Equals(this.Content, other.Content)
+                && object.Equals(this.Name, other.Name)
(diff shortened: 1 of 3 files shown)


CSharp/Library/Microsoft.Bot.Connector.Shared/AttachmentEx.cs
```diff
@@ -0,0 +1,34 @@
...
+        public override bool Equals(object other)
+        {
+            return this.Equals(other as Attachment);
+        }
+
+        public override int GetHashCode()
+        {
+            var code = this.ContentType.GetHashCode()
+                ^ this.ContentUrl.GetHashCode()
+                ^ (this.Content == null ? 13 : this.Content.GetHashCode())
```
msft-shahins: why 13, 17, and 23?

EricDahlvang: @msft-shahins these fields could be null.  Prime numbers ensure if any combinatuon of two fields are null, the object will not be equal to a different instance having other two fields null. http://stackoverflow.com/questions/263400/what-is-the-best-algorithm-for-an-overridden-system-object-gethashcode

EricDahlvang: Another example is like this:

```
int hash = 13;
hash = (hash * 7) + field1.GetHashCode();
hash = (hash * 7) + field2.GetHashCode();
...
return hash;
```

from here: http://stackoverflow.com/questions/371328/why-is-it-important-to-override-gethashcode-when-equals-method-is-overridden?rq=1

But, I suppose they could be any arbitrary number since we are doing an xor.

PR #2694, opened 2017-05-03
[C# SDK] Speech extensions for Cards

brandonh-msft:
After working with a few partners to enable speech on their bots, I've created a set of extension methods that worked really well for them which I felt might aid other bot developers in creating rich bots that work well on speech.

**The goal of these changes is to reduce the friction in creating speech responses for bots utilizing rich cards today**

Here's a list & explanation of the changes:
### GetSpeech()
This is an extension method off `Attachment` and all Card objects. The goal of this method is to extract *some* useful speech response, in SSML, that can be used as a first-pass for enabling speech in a rich-interfaced bot.
The approach is implemented based on the type of card represented by the Attachment, and defined as follows:
* `ReceiptCard`: Output: "[title] `[pause]` includes [count] item(s) `[pause]` Your total is [total]"
* `SigninCard`: Output: "[title] `[pause]` [spoken choices]"
* All other cards: Output: "[title] `[pause]` [subtitle] `[pause]` [text] [spoken choices]" if any of these are empty, they are skipped.
*see below for details on spoken choices*

### GetOptionSpeech()
Able to be run on a message, this takes the `SuggestedActions.Actions` collection, if present, or the set of buttons on the first card in the `.Attachments` collection and turns them in to spoken choices. See details on spoken choices below for and outline of what that sounds like

### AsOptionsSsml()
**Spoken Choices**
Consumers can use this method off an `IEnumerable<CardAction>` to get spoken choices. Additionally, this is used by the methods mentioned above to speak available choices back to the end user. 
These are spoken as "You can say Item 1 [pause] Item 2 [pause], or Item 3" in the case of >2 items or
"You can say Item 1 or Item 2" in the case of 2 items or
"You can say Item 1" in the case of a single item

### GetSpeechForCarousel()
Combines the output of `GetSpeech()` above with the language logic from `AsOptionsSsml` to read out multiple cards in a response.

### CombineSsml()
This method is intended to be used by a developer if they get output from more than 1 of the above-mentioned methods. They can take the output from one and add it in to the end of the SSML output from another.
eg: `GetSpeechForCarousel().CombineSsml(msg.GetOptionSpeech())` would read out a carousel, then present the user with "you can say ..." option choices for the `SuggestedActions` collection on the message.

### DIFF
--- a/CSharp/Library/Microsoft.Bot.Connector.Shared/CardEx.cs
+++ b/CSharp/Library/Microsoft.Bot.Connector.Shared/CardEx.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
+using System.Xml.Linq;
 
 namespace Microsoft.Bot.Connector
 {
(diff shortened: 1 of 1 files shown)


brandonh-msft:
note i had accidently done this work on top of the wrong branch; the "node SDK" commits are not in the latest `push` (sorry for the confusion)

CSharp/Library/Microsoft.Bot.Connector.Shared/CardEx.cs
```diff
@@ -87,5 +88,276 @@ private static Attachment CreateAttachment<T>(T card, string contentType)
...
+            {
+                return GetSpeech(attachment.Content as ReceiptCard);
+            }
+            else if (attachment.Content is SigninCard)
+            {
+                return GetSpeech(attachment.Content as SigninCard);
+            }
+            else
+            {
+                return GetStandardCardSpeech((dynamic)(attachment.Content));
```
willportnoy: >dynamic [](start = 46, length = 7)

cute trick for dynamic dispatch - are there that many cases?

CSharp/Library/Microsoft.Bot.Connector.Shared/CardEx.cs
```diff
@@ -87,5 +88,276 @@ private static Attachment CreateAttachment<T>(T card, string contentType)
...
+        }
+        /// <summary>
+        /// Gets SSML markup for this <see cref="ReceiptCard"/>.
+        /// </summary>
+        /// <param name="receiptCard"></param>
+        /// <returns></returns>
+        public static string GetSpeech(this ReceiptCard receiptCard)
+        {
+
+            var title = receiptCard?.Title;
```
willportnoy: >? [](start = 35, length = 1)

it's not clear to sprinkle the coalesce operator everywhere - either it's null or not

CSharp/Library/Microsoft.Bot.Connector.Shared/CardEx.cs
brandonh-msft: If the cards all inherited from a common type we wouldn't have to, so yeah. Audio, Video, Hero, Thumbnail, Animation. All have same structure, but no common type.

CSharp/Library/Microsoft.Bot.Connector.Shared/CardEx.cs
brandonh-msft: Can blow out all the nullable operator usage, but code's going to get quite verbose. go for it?

brandonh-msft:
⏲

PR #2697, opened 2017-05-04
[C# SDK] PostAsync(string) should set the InputHint to ExpectingInput 

brandonh-msft:
I've yet to see a situation where a bot was using PostAsync(string) and didn't expect the conversation to continue from the user in some fashion.
When this isn't the case, we'd expect the bot to be sending multiple messages, ending a conversation, etc where the developer is more likely to be using `context.MakeMessage()` to send full message objects.

### DIFF
--- a/CSharp/Library/Microsoft.Bot.Builder/Dialogs/IDialogContext.cs
+++ b/CSharp/Library/Microsoft.Bot.Builder/Dialogs/IDialogContext.cs
@@ -31,11 +31,11 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 //
 
+using Microsoft.Bot.Builder.Dialogs.Internals;
+using Microsoft.Bot.Connector;
 using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.Bot.Builder.Dialogs.Internals;
-using Microsoft.Bot.Connector;
 
 namespace Microsoft.Bot.Builder.Dialogs
 {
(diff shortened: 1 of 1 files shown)


brandonh-msft:
Optionally, I wouldn't hate it if the `PostAsync(string)` extension took `string inputHint = InputHints.ExpectingInput` and set `.InputHint = inputHint`
Totally willing to redo in that way if we're open to it.

CSharp/Library/Microsoft.Bot.Builder/Dialogs/IDialogContext.cs
```diff
@@ -141,6 +141,7 @@ public static partial class Extensions
         {
             var message = botToUser.MakeMessage();
             message.Text = text;
+            message.InputHint = InputHints.ExpectingInput;
```
msft-shahins: AutoInputHint_BotToUser will automatically apply InputHints to messages that don't have InputHint: https://github.com/Microsoft/BotBuilder/blob/master/CSharp/Library/Microsoft.Bot.Builder/ConnectorEx/BotToUser.cs#L182

It is wrong to set the InputHint to expecting for the cases that user might use the following PostAysnc: https://github.com/Microsoft/BotBuilder/blob/master/CSharp/Library/Microsoft.Bot.Builder/Dialogs/IDialogContext.cs#L142 to send multiple messages to channel.

brandonh-msft: I agree that in the case of multiple messages on the stack we wouldn't want to do this. Let's make it so it only happens in the case where there's a single message on the stack. Can you help with that?

PR #2707, opened 2017-05-05
[Node SDK] add 2 new EntityRecognizer methods for new LUIS List Entities

stevengum97:
LUIS's List Entities return in a slightly different format than normal entities, they have an additional `"resolution"` property which contains another key-value pair: the key `"values"` and an array of strings. These strings can be construed as the true value a developer is looking for, opposed to the List Entity name. 

The methods are `findListEntity()` and `findAllListEntities()`.
 
This is a link to the current API docs regarding [**List Entities**](https://dev.projectoxford.ai/docs/services/5890b47c39e2bb17b84a55ff/operations/5890b47c39e2bb052c5b9c14).

##### Example List Entity from LUIS:

    {
      "entity": "ice cream",
      "type": "Desserts",
      "startIndex": 12,
      "endIndex": 20,
      "resolution": {
          "values": [
              "FrozenDesserts"
          ]
      }
    }

### DIFF
--- a/Node/core/lib/dialogs/EntityRecognizer.js
+++ b/Node/core/lib/dialogs/EntityRecognizer.js
@@ -13,11 +13,36 @@ var EntityRecognizer = (function () {
         }
         return null;
     };
+    EntityRecognizer.findListEntity = function (entities, value) {
+        for (var i = 0; entities && i < entities.length; i++) {
+            if (entities[i].resolution) {
+                for (var j = 0; j < entities[i].resolution.values.length; j++) {
+                    if (entities[i].resolution.values[j] == value) {
+                        return entities[i];
+                    }
+                }
+            }
+        }
+        return null;
+    };
     EntityRecognizer.findAllEntities = function (entities, type) {
         var found = [];
         for (var i = 0; entities && i < entities.length; i++) {
             if (entities[i].type == type) {
                 found.push(entities[i]);
+            }
+        }
+        return found;
+    };
+    EntityRecognizer.findAllListEntities = function (entities, value) {
+        var found = [];
+        for (var i = 0; entities && i < entities.length; i++) {
+            if (entities[i].resolution) {
+                for (var j = 0; j < entities[i].resolution.values.length; j++) {
+                    if (entities[i].resolution.values[j] == value) {
+                        found.push(entities[i]);
+                    }
+                }
             }
         }
         return found;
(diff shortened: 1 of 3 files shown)


Stevenic:
Hi Steven... I'm mulling this over. Can you provide a couple of code examples for how you'd use these methods in your bot? Closed lists are a nice add to LUIS but semantically they are the same as a normal entity, they've just had their value validated in much the same way a date has been validated and therefore has resolution.

My pushback is that if you have a list entity with 50 possible canonical values I don't see you making 50 separate calls looking for each canonical value. Even in your example I think you'd look for the "Desserts" entity and then use the resolution value instead of what the user typed. So I think you'd still work with the entity in much the same way you do any other entity.

stevengum97:
@Stevenic you're right, for large List Entities `findListEntity()` and `findAllListEntities()` are not going to be worth it to call these methods 50 times to look for List Entities. However, I think that might not be the normal use case for List Entities (I could be wrong). Immediately I have only two ideas for potential entities at that size, the States (which is provided as the example in the API Docs) and countries (much in the same capacity of the States).

I think it might be of use to developers who know exactly what they're looking for in an utterance from a user. The `findListEntity()` method abstracts some of the code a dev would have to write if the dev is looking for a specific cannonical form and/or has a small List Entity (which I think might be closer to the norm). I think that if the user is trained or primed to provide utterances that use list entities already, the `findListEntity()` methods might be more beneficial/faster to write bots with.


    // Code for tests case where List Entities are small
    // Example Utterance: "I need cable help in WA."

    function (session, args, next) {
        var state = builder.EntityRecognizer.findEntity(args.entities, 'State');
        if (state.resolution && state.resolution.values) {
            for (var i = 0; i < state.resolution.values.length; i++) {
                if (state.resolution.values[i] === ('Washington' || 'Oregon')) {
                    state = state.resolution.values[i];
                }
            }
            if (!state) {
            session.send('We only offer support in Washington and Oregon');
            builder.Prompts.choice(session, 'Please choose one:', ' Washington | Oregon');
            } else {
                next({ state: state });
            }
        }
    }

#### VS
    
    function (session, args, next) {
        var stateOrcinus = builder.EntityRecognizer.findListEntity(args.entities, 'Washington');
        var stateCastor = builder.EntityRecognizer.findListEntity(args.entities, 'Oregon');
        if ( !(stateOrcinus||stateCastor) ) {
            session.send('We only offer support in Washington and Oregon');
            builder.Prompts.choice(session, 'Please choose one:', ' Washington | Oregon');
       } else {
            next({ state: stateOrcinus || stateCastor });
       }
    }

That said I think placing this on hold for now is probably the best idea; if it comes up as a common request it can be revisited.

stevengum97:
Closing PR.

PR #2718, opened 2017-05-08
[C# SDK] Handling null text messages for LuisDialog

q3blend:
LuisDialog throws exception if the message text is null.

I am mapping this to the empty string intent.

### DIFF
--- a/CSharp/Library/Microsoft.Bot.Builder/Dialogs/LuisDialog.cs
+++ b/CSharp/Library/Microsoft.Bot.Builder/Dialogs/LuisDialog.cs
@@ -213,7 +213,7 @@ namespace Microsoft.Bot.Builder.Dialogs
 
             if (winner == null)
             {
-                throw new InvalidOperationException("No winning intent selected from Luis results.");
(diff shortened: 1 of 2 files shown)


willportnoy:
https://github.com/Microsoft/BotBuilder/issues/2871 seems related

willportnoy:
It doesn't seem reasonable to default to the none intent when invalid data is passed to the LuisDialog - I'm happy to accept a pull request that throws a more meaningful exception though.

q3blend:
@willportnoy I agree with you that this is not a good default behavior, but I don't think sending a more meaningful exception would solve the issue. We need something to stop LuisDialog from breaking on null or empty string messages. Maybe a default NullMessageHandler that could be overridden?

willportnoy:
If you want to add special handling for null messages, you can add a global scorable like

https://github.com/Microsoft/BotBuilder/blob/3c912ec90ad08a63e7ffd2256998db3521b028aa/CSharp/Tests/Microsoft.Bot.Builder.Tests/ScorableDialogTests.cs#L100

PR #2723, opened 2017-05-09
[Node SDK]  Add French localisation default phrases

simonpo:
Added localised numbers, ordinals and a couple of reverse_ordinals for /fr


nwhitmont:
approved @Stevenic

PR #2735, opened 2017-05-09
[Node SDK] Modify LuisRecognizer for CompositeEntities

stevengum97:
modified recognize() and created CompositeEntities interfaces


stevengum97:
Added new LUIS interfaces for recognizer updates. Deleted RecognizerFilter.js

PR #2749, opened 2017-05-11
[Node SDK] Fix for #1530 - add check for NaN before attempting to recognize number

nwhitmont:
add check for NaN before attempting to recognize number input from Prompt.number()


nwhitmont:
fix for #1530

PR #2751, opened 2017-05-11
[Node SDK] Add Turkish (tr) localization default phrases

ibrahimonat:


Stevenic:
If you look at the [English version](https://github.com/Microsoft/BotBuilder/blob/master/Node/core/lib/locale/en/BotBuilder.json) of BotBuilder.json there are some additional entries that should be localized. Can we get you to add Turkish versions of these additional entries?

PR #2346, opened 2017-02-25
Enable usage of Bot.Builder with ASP.Net Core

iedeny:
### Purpose of changes

The goal of proposed change is to decouple **Bot.Builder** projects from specific version of **Bot.Connector** that will enable us to use **Bot.Builder** with **ASP.Net Core**.
Currently it is not possible due to following dependency chain: **Bot.Builder** -> **Bot.Connecor** -> “regular” **ASP.Net** 4.6 (see image below).

Dependency diagram before changes:
vs codemap original

Dependency diagram after proposed changes:
vs codemap proposed

### Project level changes:
-	Create new portable *Bot.Connector.Common.csproj* based on *Bot.Connector.Shared.shproj*.
	New project will  that target **.NET Standard** 1.4 which can be referenced from both **.NET Framework** and **.NET Core** ([details](https://blogs.msdn.microsoft.com/dotnet/2016/09/26/introducing-net-standard/))
-	Change *Bot.Connector.AspNetCore* to target **.NET Standard** 1.4 (currently targets 1.6). 
	This will allow *Bot.Connector.AspNetCore* to be referenced by **.NET Framework** 4.6.1 and **.NET Core** 1.0 projects
	(currently there is no version of **.NET Framework** that implemented **.NET Standard** 1.6)
-	Create new *Bot.Connector.AspNetCore.Mvc* projects which contain *TrustServiceUrlAttribute*
    -   *Bot.Connector.AspNetCore.Mvc* targets **.NET Standard** 1.6 (we cannot target 1.4 because [Microsoft.AspNetCore.Mvc](https://www.nuget.org/packages/Microsoft.AspNetCore.Mvc/)
		has a dependency on **.NET Standard** 1.6). As a result currently this project can only be used with **.NET Core**. 
		This limitation should go away once **.NET Standard** 2.0 is released and supported by MVC.
    -   *Bot.Connector.AspNetCore.Mvc.NetFramework* targets **.NET Framework** 4.6.1. This porject should be used if you are [using ASP.Net Core with **.NET Framework**](http://www.hanselman.com/blog/content/binary/Windows-Live-Writer/ASP.Net-5-is-dead---Introducing-ASP.NE.0_E068/image_60140ec0-ce46-4dbf-a14f-4210eab7f42c.png). 
    -   Once **.NET Standard** 2.0 is released and supported by MVC we can change first project to target it. As a result second project can be removed.
    -   Both projects are distributed as a single package (depending on platform nuget will reference one assembly or another)
[...]

PR #2761, opened 2017-05-14
Utilise new csproj format to consolidate Microsoft.Bot.Connector

damienpontifex:
- Multiple TargetFrameworks in one project allowing consolidation of net45 and netstandard targets in one project
- Using csproj as nuget definition allows removal of nuspec file

There are a lot of renames which is just moving the shared project into the one location for this project.

n.b. I didn't update the cmd scripts for client generation as I wasn't sure where these are executed. They would move to be `dotnet pack` commands from the appropriate csproj locations


msft-shahins:
Can you update the PR and use develop branch that is already have some consolidated net core support?

damienpontifex:
hmm quite a few conflicts with that. I'll close and submit another PR with changes against develop branch.

msft-shahins:
here is the pull request describing current structure of libraries in develop branch: https://github.com/Microsoft/BotBuilder/pull/2346.

Issue #2778, opened 2017-05-17
[Known Issue] Prompts.confirm doesn't work properly in Chinese (zh-Hans)


josephktcheung:
## Pre-flight checklist
- [x] I have [searched the existing issues](https://github.com/Microsoft/BotBuilder/issues?utf8=%E2%9C%93&q=is%3Aissue) before logging a new one.
---------------------------

## System Information (Required)

* SDK Pletform: Node
* SDK Version: 3.8.3
* Development Environment: localhost

## Issue Description

Botbuilder cannot recognise "是" as 1 of the yes expressions.

## Example Code

```javascript
const {
  ConsoleConnector,
  UniversalBot,
  Prompts,
} = require('botbuilder');

const connector = new ConsoleConnector().listen();
const bot = new UniversalBot(connector,{
  localizerSettings: {
    defaultLocale: 'en',
  },
});

bot.dialog('/', [
  (session) => {
    session.preferredLocale('zh-Hans', () => {
      Prompts.confirm(session, 'Confirmation');
    });
  },
  (session, results) => {
    console.log('results? ', results);
    session.endDialog();
  },
]);
```

## Steps to Reproduce
1. Start bot
2. Enter "是" when "Confirmation" shows up

## Expected Behavior

Console should show log as below:

```javascript
results?  { resumed: 0,
  response: true,
  childId: 'BotBuilder:prompt-confirm' }
```

## Actual Results

Console shows "我没听明白。 请选择'是'或'否'。", meaning Botbuilder doesn't recognise `是` as 1 of the `yes` expressions.

After investigation, the localisation of `boolean_choices`, `yesExp` and `noExp` in zh-Hans' [BotBuilder.json](https://github.com/Microsoft/BotBuilder/blob/master/Node/core/lib/locale/zh-Hans/BotBuilder.json#L13) are incorrect, those unicode characters have 1 more `\` than needed e.g. `\\u662f` but the correct unicode should be `\u662f` for chinese character `是`, which makes BotBuilder unable to recognise Chinese character `是` as 1 of the `yes` expressions.

nwhitmont:
@josephktcheung Thanks for reporting. This is a known issue & we're working to get a patch out to fix it soon.

nwhitmont:
Moved to internal issue tracker.

PR #2779, opened 2017-05-17
[Node SDK] Fixed Chinese localization regex patterns

josephktcheung:
Make BotBuilder recognize Chinese characters for boolean expressions correctly.

Referencing issue #2778


Stevenic:
Thanks

Issue #2561, opened 2017-04-06
[Feature Request] NodeJS Calling SDK should allow for clock skew when verifying auth token


ctstone:
The Node SDK validates the JWT token twice, but only on the second validation does it take clock skew into account. See PR #2502 for the fix and see Issue #1885 for the original issue that was closed.

nwhitmont:
Thanks for reporting this @ctstone - I have forwarded this to the Bot Framework Node SDK Platform Team.

ctstone:
@nwhitmont, this is now causing random bot hang ups when running on Azure App Services. It wasn't a big deal when it only affected my local dev environment, but if AAD and Azure App Services are not in sync, this is a bigger problem.

sgellock:
Thank you for opening an issue against the Bot Framework SDK v3. As part of the Bot Framework v4 release, we’ve moved all v3 work to a new repo located at https://github.com/microsoft/botbuilder-v3. We will continue to support and offer maintenance updates to v3 via this new repo.
 
From now on, https://github.com/microsoft/botbuilder repo will be used as hub, with pointers to all the different SDK languages, tools and samples repos.
 
As part of this restructuring, we are closing all tickets in this repo.
 
For defects or feature requests, please create a new issue in the new Bot Framework v3 repo found here:
https://github.com/microsoft/botbuilder-v3/issues

For Azure Bot Service Channel specific defects or feature requests (e.g. Facebook, Twilio, Teams, Slack, etc.), please create a new issue in the new Bot Framework Channel repo found here:
https://github.com/microsoft/botbuilder-channels/issues
 
For product behavior, how-to, or general understanding questions, please use Stackoverflow.
https://stackoverflow.com/search?q=bot+framework


Thank you.
 
The Bot Framework Team

PR #2784, opened 2017-05-17
[Node SDK] remove duplicate jwt check. fix for #2561

ctstone:
This PR removes the duplicate JWT verification and relies solely on the implementation in `jsonwebtoken` using the already existing `jwtVerifyOptions`

I also added a call to `console.error` when the connector encounters a token error, in order to provide feedback for otherwise mysterious hangups.

CallConnector.js was built using tsc 2.3.2


Stevenic:
Thanks

Issue #2780, opened 2017-05-17
[Code Examples] There might be a bug in AnnotatedSandwichBot sample


CXuesong:
## Pre-flight checklist
- [ ] I posted a question about this issue on [Stack Overflow](http://stackoverflow.com/search?q=%23botframework) using the tag `botframework`.
- [x] I have [searched the existing issues](https://github.com/Microsoft/BotBuilder/issues?utf8=%E2%9C%93&q=is%3Aissue) before logging a new one.
---------------------------

## System Information (Required)

* SDK Platform: .NET
* SDK Version: 3.8.3 (master)
* Development Environment: localhost

## Issue Description

At the step "select one or more toppings", if a user chooses No. 13, i.e. "No Preference", a `FormCanceledException` caused by `ArgumentNullException` will be thrown.

## Example Code

The `ArgumentNullException` is caused by https://github.com/Microsoft/BotBuilder/blob/master/CSharp/Samples/AnnotatedSandwichBot/AnnotatedSandwich.cs#L120.

The `value` parameter is `null` for "No Preference", but at L120, it's assumed that `value` cannot be `null`.

## Steps to Reproduce
1. `git clone https://github.com/Microsoft/BotBuilder`
2. Open the solution in VS 2017.
3. Choose` Microsoft.Bot.Sample.AnnotatedSandwichBot` as startup project.
4. Build & run the project.
5. Connect to the end point using emulator.
6. Trigger the first question, then chooses any option again and again, until the bot says "Please select one or more toppings".
7. Enter `13` .

## Expected Behavior

"For sandwich toppings you have selected (something)"

## Actual Results

The bot got short-circuited.
![default](https://cloud.githubusercontent.com/assets/11611687/26145642/bd0ede76-3b1f-11e7-8334-3f78c23905f9.PNG)

```
{
  "type": "message",
  "timestamp": "2017-05-17T08:37:16.503Z",
  "localTimestamp": "2017-05-17T16:37:16+08:00",
  "serviceUrl": "http://127.0.0.1:53953",
  "channelId": "emulator",
  "from": {
    "id": "e9bddmhaff99ch7n6",
    "name": "Bot"
  },
  "conversation": {
    "id": "62e120g0effdi956dc"
  },
  "recipient": {
    "id": "default-user"
  },
  "membersAdded": [],
  "membersRemoved": [],
  "locale": "en",
  "text": "Exception: Value cannot be null.\r\nParameter name: source",
  "attachments": [
    {
[...]

PR #2786, opened 2017-05-18
[.NET] Code Examples - Proposed fix for #2780.

CXuesong:
Ref = #2780 
This will fix the short-circuit problem for "No Preference" in topping selection.


JasonSowers:
Thank you for your submission

PR #2831, opened 2017-05-25
Using modern project format for shared projects and aspnet core

damienpontifex:
- Moving nuget properties into project to utilise dotnet pack and one definition for project and package
- Moving projects to use new project and dependency definition
- Removed specific mvc projects for core and full framework and utilised multiple TFM properties in TargetFrameworks
- Moved aspnet core mvc dependency to Microsoft.AspNetCore.Mvc.Core instead of Microsoft.AspNetCore.Mvc as we only need enough to host controllers, not render views etc

Follow up from #2761 with @msft-shahins with PR against develop branch

_n.b. createpackage.cmd and `dotnet pack` still needs discussion for setting version property. This would be done in csproj definition and can use --version-suffix parameter for other releases if needed_


damienpontifex:
@msft-shahins any update/feedback on this?

PR #2851, opened 2017-05-29
[C# SDK] Allow partial matches when recognizing choices in a PromptChoice

ejadib:
@msft-shahins FYI

Related to https://stackoverflow.com/questions/44241355/botframework-promptdialog-choice-no-longer-returns-partial-matches


hpksp:
Seems like this feature hasn't been resolved yet.

Issue #2839, opened 2017-05-26
[LUIS] triggerAction matching LuisRecognizer intent cannot interrupt current dialog


josephktcheung:
## Pre-flight checklist
- [x] I have [searched the existing issues](https://github.com/Microsoft/BotBuilder/issues?utf8=%E2%9C%93&q=is%3Aissue) before logging a new one.
---------------------------

## System Information (Required)
* SDK Platform: Node.js
* SDK Version: 3.8.3
* Development Environment: localhost

## Issue Description
`triggerAction` which matches LuisRecognizer intent cannot interrupt current dialog e.g. choice.

## Example Code
I modified `Node/examples/basics-naturalLanguage` code (btw the [current code in master branch](https://github.com/Microsoft/BotBuilder/blob/master/Node/examples/basics-naturalLanguage/app.js#L21) doesn't work anymore because the luis url returns `{ "statusCode": 404, "message": "Resource not found" }`, I should probably create another issue for this).

```javascript
var builder = require('../../core/');

// Create bot and bind to console
var connector = new builder.ConsoleConnector().listen();
var bot = new builder.UniversalBot(connector, function (session) {
    session.send("Hi... I'm a bot sample that when I'm asking for a choice I cannot be interrupted by user intent.");
});

// Add global LUIS recognizer to bot
var model = process.env.model || 'https://westus.api.cognitive.microsoft.com/luis/v2.0/apps/b24dac03-c682-4e41-be69-c56513dcd748?subscription-key=76b1f832aa0c4e589bead8241e06aac0';
bot.recognizer(new builder.LuisRecognizer(model));

bot.dialog('/hello', [
    function (session, args, next) {
        builder.Prompts.choice(session, 'what do you want?', 'pizza|whiskey|water')
    },
    function (session, results, next) {
        switch (results.response) {
            case 'pizza':
                session.endDialog('I love pizza as well');
                break;
            case 'whiskey':
                session.endDialog('Which brand do you like?');
                break;
            case 'water':
                session.endDialog('Good taste');
                break;
            default:
                session.endDialog();
                break;
        }
    }
]).triggerAction({
    matches: 'hello',
});

bot.dialog('/bye', [
    function (session, args, next) {
        session.endDialog('see you soon');
    }
]).triggerAction({
    matches: 'bye'
});
```

## Steps to Reproduce
1. type `hi`
2. console shows `what do you want? (1. pizza, 2. whiskey, or 3. water)`
3. type `bye`
[...]

PR #2859, opened 2017-05-31
[Node SDK] Fix Prompt IIntentRecognizerResult

josephktcheung:
Fix Prompt dialog that cannot be interrupted by global action with LUIS intent.
Referencing issue #2839

### DIFF
--- a/Node/core/lib/dialogs/Prompt.js
+++ b/Node/core/lib/dialogs/Prompt.js
@@ -105,7 +105,7 @@ var Prompt = (function (_super) {
         }
         var idx = 0;
         var handlers = this._onRecognize;
-        var result = context.intent || { score: 0.0, intent: null };
+        var result = { score: 0.0, intent: null };
         function next() {
             try {
                 if (idx < handlers.length) {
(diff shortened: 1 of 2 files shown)


Stevenic:
Good catch...

singzzz:
This change compromises my implementation of LUIS and QnA maker. Everything (almost) now gets matched to the qna sentiment, including typos:

var recognizer = new builder.LuisRecognizer(LuisModelUrl);
bot.recognizer(recognizer);
bot.recognizer(qnaRecognizer);

 if (!session.toilet.component) {
            session.beginDialog('/toiletComponent', component, args);
        }

eg: What part of the toilet is broken? (1. Cistern, 2. Pipe, or 3. Seat)

Anything except an exact match gets pickedup by qna sentiment which replaces the dialog stack

Is there a better implementation I should be using, other than adding 'context.intent ||' back into prompt.js?

Thanks

EricDahlvang:
Hi @singzzz 

You can disable the recognizer while within a dialog with something like:

```js
var recognizer = new builder.LuisRecognizer(LuisModelUrl);
    .onEnabled(function (context, callback) {
        var enabled = context.dialogStack().length === 0; 
        callback(null, enabled); 
    });
```

If you require further assistance, please ask a question on Stack Overflow: https://stackoverflow.com/questions/tagged/botframework

singzzz:
Thanks Eric, but I cant get this to work for the qna recognizer, will post on stack overflow as suggested.

PR #2879, opened 2017-06-02
Refactor LuisModelAttribute to support third-party implementations

### LINKED ISSUE #2230: [LUIS] ILuisModel difficult to inherit from
We're attempting to upgrade `Microsoft.Bot.Builder` from `3.2.1` to `3.5.1` and we've started getting build errors. Our custom implementation of `ILuisModel` is missing two new properties: `ApiVersion` and `UriBase`.

`ApiVersion` is easy to add; `UriBase` however requires referencing the framework source to reimplement:

```csharp
public LuisApiVersion ApiVersion => LuisApiVersion.V2;
public Uri UriBase => LuisModelAttribute.LuisEndpoints[ApiVersion];
```

I'm wondering if there'd be an easier way to expose the `LuisEndpoints` values. Right now, I'm thinking we could use an `Attribute` (perhaps the built-in `DescriptionAttribute`) on the `LuisApiVersion` enum values. We could also move the logic for generating the `UriBase` up to the `LuisService` layer, but that decreases flexibility.

Let me know your thoughts! I'd be happy to implement this change 😄

cglong:
Fixes #2230. Basically, `LuisModelAttribute` contained a few domain-specific pieces of logic that prevented third parties from consuming their own implementations of `ILuisModel`. This is a small refactoring that brings some of that logic to the BF framework itself:

* Bring optional LUIS properties to `ILuisModel`
* Calculate `UriBase` in `LuisRequest`
* Move `ILuisModel.ModifyRequest` to `LuisService`

### DIFF
--- a/CSharp/Library/Microsoft.Bot.Builder/Luis/LuisModel.cs
+++ b/CSharp/Library/Microsoft.Bot.Builder/Luis/LuisModel.cs
@@ -52,6 +52,12 @@ namespace Microsoft.Bot.Builder.Luis
     /// </summary>
     public interface ILuisModel
     {
+/// <summary>
+        /// Domain where LUIS application is located.
+        /// </summary>
+        /// <remarks>Null means default which is api.projectoxford.ai for V1 API and westus.api.cognitive.microsoft.com for V2 api.</remarks>
+        string Domain { get; }
+
         /// <summary>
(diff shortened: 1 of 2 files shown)


willportnoy:
I'm going to close this as we worked on an updated solution together.  Feel free to ping me.

PR #2953, opened 2017-06-15
[Node SDK] Add fallback for locales in LuisRecognizer model

spghassenrjab:
If a user's locale is "en-US" and a developer specified just "en" locale in LuisRecognizer model, we use "en" locale as a fallback

### DIFF
--- a/Node/core/lib/dialogs/LuisRecognizer.js
+++ b/Node/core/lib/dialogs/LuisRecognizer.js
@@ -31,6 +31,13 @@ var LuisRecognizer = (function (_super) {
             var utterance = context.message.text;
             var locale = context.locale || '*';
             var model = this.models.hasOwnProperty(locale) ? this.models[locale] : this.models['*'];
+            if (!model) {
+                var dashIndex = locale.indexOf('-')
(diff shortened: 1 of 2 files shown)


stevengum97:
Can you add this to LuisRecognizer.ts and then transpile it? Thanks.

Stevenic:
This is a great catch but I've changed your code to cover the case where you might have both a '*" recognizer and a 'en' recognizer. Your code would have fallen back to '*' in that case and still not used the 'en' one.

spghassenrjab:
You are right. I made the change too in LuisRecognier.ts file in line 60, the instruction becomes
`var model = this.models[locale];`
instead of
`var model = this.models.hasOwnProperty(locale) ? this.models[locale] : this.models['*'];`
I transpiled and pushed the change but the pull request is already closed.

spghassenrjab:
This is a friendly reminder as a [commit](https://github.com/Microsoft/BotBuilder/commit/8db32849d9da38a06e0fd1ff11f8e8244fa3a8cc) have been created based on my PR.
I would have liked some credits next time. I know it's not much, but it was important to me.

PR #2982, opened 2017-06-19
[Node] LuisRecognizer: remove mentions from recognized text

suyash:
On Microsoft Teams, there are 2 scope of interaction with a bot, a team scope (in a channel) and a personal scope (1:1) conversation. In a team scope, to send a message to a bot, it needs to be explicitly invoked with a `@`, but the text sent to the framework contains the invocation name inside `<at>` tags, which when sent to the model to recognize returns incorrect results.

This tries to fix that by removing all occurrences of `<at>...</at>` from the message before querying LUIS for recognition.

### DIFF
--- a/Node/core/lib/dialogs/LuisRecognizer.js
+++ b/Node/core/lib/dialogs/LuisRecognizer.js
@@ -28,7 +28,9 @@ var LuisRecognizer = (function (_super) {
     LuisRecognizer.prototype.onRecognize = function (context, callback) {
         var result = { score: 0.0, intent: null };
         if (context && context.message && context.message.text) {
-            var utterance = context.message.text.replace(/<at>.*<\/at>/g, '');
+            var utterance = context.message.text
+                .replace(/<at>.*<\/at>/g, '')
+                .replace(/@.*?(\s|$)/g, '');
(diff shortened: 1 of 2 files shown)


Node/core/lib/dialogs/LuisRecognizer.js
```diff
@@ -28,7 +28,7 @@ var LuisRecognizer = (function (_super) {
     LuisRecognizer.prototype.onRecognize = function (context, callback) {
         var result = { score: 0.0, intent: null };
         if (context && context.message && context.message.text) {
-            var utterance = context.message.text;
+            var utterance = context.message.text.replace(/<at>.*<\/at>/g, '');
```
stevengum97: This should also include other channels, and other methods of mentioning users and bots, like @

suyash:
I have added code to replace words beginning with `@`. I am mainly familiar with teams, is the number of ways of mentioning a user in a message documented somewhere

stevengum:
Closing as this is out of date and the code no longer lives here but is instead in [botbuilder-v3](https://github.com/microsoft/botbuilder-v3). There was also a differing of opinions on where this code should live.

PR #3007, opened 2017-06-22
[Node] Handle LUIS HTTP errors

greuze:
Handle LUIS HTTP errors, so the bot can know that anything went wrong. For example if the URL has `http` protocol instead of `https`, the bot gets a 404, if the subscription key is wrong, the bot gets a 401, but in both cases, it thinks that everything went ok, with a result `{ score: 0, intent: null, intents: [], entities: [] }`

### DIFF
--- a/Node/core/lib/dialogs/LuisRecognizer.js
+++ b/Node/core/lib/dialogs/LuisRecognizer.js
@@ -83,22 +83,20 @@ var LuisRecognizer = (function (_super) {
             request.get(url.format(uri), function (err, res, body) {
                 var result;
                 try {
-                    if (!err) {
-                        if (res.statusCode === 200) {
-                            result = JSON.parse(body);
-                            result.intents = result.intents || [];
-                            result.entities = result.entities || [];
-                            result.compositeEntities = result.compositeEntities || [];
-                            if (result.topScoringIntent && result.intents.length == 0) {
-                                result.intents.push(result.topScoringIntent);
-                            }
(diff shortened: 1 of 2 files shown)


stevengum97:
Can you transpile your code please? Thanks

greuze:
@stevengum97 Done. Please, note that the transpiled file was not updated, so changes to `compositeEntities` appeared in the `.js` file too.

greuze:
I fixed the conflicts with master, and rebased the code, in order to be easier to see the changes.

stevengum97:
Can you change your code to the following and then I'll accept it. Thanks again for your catch.

```ts
try {
    if (res && res.statusCode === 200) {
        if (!err) {
            result = JSON.parse(body);
            result.intents = result.intents || [];
            result.entities = result.entities || [];
            result.compositeEntities = result.compositeEntities || [];
            if (result.topScoringIntent && result.intents.length == 0) {
                result.intents.push(result.topScoringIntent);
            }
            if (result.intents.length == 1 && typeof result.intents[0].score !== 'number') {
                // Intents for the builtin Cortana app don't return a score.
                result.intents[0].score = 1.0;
            }
        }
    } else {
        err = new Error(body);
    }
} catch (e) {
    err = e;
}
```

greuze:
@stevengum97 Done. I removed the `if (!err)` as it is redundant after checking `if (res)`, they both will never be defined at the same time (cannot be a timeout or DNS resolution error after a response with code 200).

stevengum97:
Thanks.

Issue #2903, opened 2017-06-08
[Question] Conversation Logger


Magic73:
I need to do some analytics with the Bot.
I'm using a TraceLogger: IActivityLogger class.

The issue is with the information that I may use. The LogAsync method, has only the activity parameter. But with this info I can't find a way to know if the activity is from the bot to the user, or from the user to the bot.

Is there a way to get this info?

Checking the BotFramework source code, I saw that the LogAsync method of the logger is called from two classes which do what I need to know:

LogPostToBot and LogBotToUser

Maybe I have to register two custom classes (which implements IPostToBot and IBotToUser)  so that the builder will use mine instead of the predefined one?
In case, how may I do it?

nwhitmont:
@Magic73 are you using .NET or Node SDK?

Magic73:
.Net

JasonSowers:
@Magic73 going to close this issue because of the PR submission.  Feel free to reopen it if you would like to continue discussing here.

Issue #2391, opened 2017-03-04
Issue with PromptDialog.Choice and Facebook's Quick Replies


brunoklein99:
Currently when using `PromptDialog.Choice` with Facebook and using `PromptStyle.Keyboard` to map the options to Quick Replies, the framework doesn't handle the returned user response message correctly, causing the prompt to repeat the message and exhaust the retries until we get `too many attemps`.

For instance, the following dialog:

```cs
[Serializable]
public class TestDialog : IDialog<object>
{
    public async Task StartAsync(IDialogContext context)
    {
        PromptDialog.Choice(
            context,
            ResumeAsync,
            new[] { "a", "b", "c" },
            "this is a prompt",
            promptStyle: PromptStyle.Keyboard,
            descriptions: new[] { "1", "2", "3" });
    }

    private async Task ResumeAsync(IDialogContext context, IAwaitable<string> result)
    {
        var message = context.MakeMessage();

        message.Text = $"You chose {await result}";

        await context.PostAsync(message);

        context.Done(new object());
    }
}
```

This dialog creates a prompt for the options, `a`, `b`, `c`, which are presented to the user as `1`, `2`, `3`. Using the web dialog from `botframework.com` we get the expected behavior.

http://i.imgur.com/x17P85E.jpg

But on Facebook, the same dialog behaves differently.

http://i.imgur.com/hsqX99j.jpg 

Messenger removes the buttons after you click them, but you get the idea.

By debugging the incoming Facebook response I could verify that the `Activity` has the proper `ChannelData` property loaded.

```
{
  "sender": {
    "id": "xxx"
  },
  "recipient": {
    "id": "xxx"
  },
  "timestamp": 1488638719106,
  "message": {
    "mid": "mid.1488444714446:4a44e6ff28",
    "seq": 19139,
    "text": "1",
    "is_echo": false,
    "quick_reply": {
      "payload": "a"
    }
  }
}
```

Maybe the framework is using the `message.text` field instead of `message.quick_reply.payload`?

Is this a bug or am I missing something?

willportnoy:
Can you share the activity object being sent from the bot framework facebook connector?  I would expect to see something that fits the `IMessageActivity` schema.

brunoklein99:
Thank you for the response:

```
{
  "type": "message",
  "timestamp": "2017-03-11T12:43:20.743006Z",
  "serviceUrl": "https://facebook.botframework.com",
  "channelId": "facebook",
  "from": {
[...]

PR #3022, opened 2017-06-23
[.NET] Update IActivityLogger.cs

Magic73:
Fix about issue #2903 on C# SDK.

https://github.com/Microsoft/BotBuilder/issues/2903

### DIFF
--- a/CSharp/Library/Microsoft.Bot.Builder/ConnectorEx/IActivityLogger.cs
+++ b/CSharp/Library/Microsoft.Bot.Builder/ConnectorEx/IActivityLogger.cs
@@ -1,4 +1,4 @@
-﻿// 
+// 
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license.
 // 
@@ -46,7 +46,7 @@ namespace Microsoft.Bot.Builder.History
     /// </summary>
     public interface IActivityLogger
     {
-        Task LogAsync(IActivity activity);
+        Task LogAsync(IActivity activity, bool fromBotToUser);
     }
 
     /// <summary>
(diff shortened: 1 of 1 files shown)


msft-shahins:
you can figure out the direction by looking at the To and Recipient address of the IActivity passed to the logger. No need to pass the flag to LogAsync function. Here is a simple implementation:

```csharp
public class Logger : IActivityLogger
{
	private readonly ConversationReference reference;
	public Logger(ConversationReference reference)
	{
		SetField.NotNull(out this.reference, nameof(reference), reference);
	}

	public async Task LogAsync(IActivity activity)
	{
		if (activity.From.Id == this.reference.Bot.Id)
		{
			Trace.WriteLine("ToUser:" + (activity as IMessageActivity)?.Text);
		}
		else
		{
			Trace.WriteLine("ToBot:" + (activity as IMessageActivity)?.Text);
		}

	}
}
```

Magic73:
But this is a channel dependant solution... it's not a solution which can be controlled by the framework.

Take this as example: https://github.com/Microsoft/BotBuilder/issues/2391
Facebook changes it's query reply implementation, and the suggested actions stops to work from a day to another.

msft-shahins:
@Magic73 the address fields are not channel specific and they are always present as required fields by the bot framework protocol. The ConversationReference created from the activity posted to bot : https://github.com/Microsoft/BotBuilder/blob/master/CSharp/Library/Microsoft.Bot.Builder.Autofac/Dialogs/DialogModule.cs#L104 will always have the bot.Id and also the activities flowing through IPostToBot and IBotToUser pipelines will have the right From and Recipient populated based on the incoming activity.

PR #3026, opened 2017-06-23
[.NET] Properties set for ILuisModel are not taken into account

shather:
After update to Bot.Builder 3.8.2:
If you create Luis Service this way:
`new LuisService(new LuisModelAttribute("id","key") { Verbose = true})`

your uri to luis won't take Verbose into account, and your url will be like:

https://westus.api.cognitive.microsoft.com/luis/v2.0/apps/x?subscription-key=y&q=query&log=True

instead of:
https://westus.api.cognitive.microsoft.com/luis/v2.0/apps/x?subscription-key=y&verbose=true&q=query&log=True

Issue #3033


willportnoy:
should this be applied to the above ```QueryAsync``` overload as well?

shather:
You are right @willportnoy .
We do not use this method and it has 0 references in Bot.Bulder solution, so i've missed it.
I've fixed that issue in : bc9fd

EricDahlvang:
@shather  Thank you for your contribution.

PR #3117, opened 2017-07-11
[Node SDK] create new object for exec in RegExpRecognizer

suyash:
Calling exec on a `RegExp` object returns null every second time for matches

screen shot 2017-07-11 at 12 09 24 pm

which makes recognizer fail every second time if the RegExps are specified as literals

```
bot.recognizer(new RegExpRecognizer('...', {
  '*': /.../,
});
```

This prevents that by creating a new RegExp object every time for exec

### DIFF
--- a/Node/core/lib/dialogs/RegExpRecognizer.js
+++ b/Node/core/lib/dialogs/RegExpRecognizer.js
@@ -31,7 +31,7 @@ var RegExpRecognizer = (function (_super) {
             var locale = context.locale || '*';
             var exp = this.expressions.hasOwnProperty(locale) ? this.expressions[locale] : this.expressions['*'];
             if (exp) {
-                var matches = exp.exec(context.message.text);
+                var matches = new RegExp(exp).exec(context.message.text);
                 if (matches && matches.length) {
                     var matched = matches[0];
                     result.score = 0.4 + ((matched.length / context.message.text.length) * 0.6);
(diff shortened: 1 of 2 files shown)


nwhitmont:
@Stevenic Actually this is expected behaviour of a global RegExp (/a/g), since in this example, @suyash is calling the regex on a single letter string ('a'), so first run, it finds the first index of 'a' **and updates properties of the regular expression object**, and second run it returns null because there is no second instance since the search (second run) was started from the `lastIndex` (aka where the first instance was found in positon 0). Thus the returned null on second run is expected when the input to the regex object is a single letter string.

To avoid this behavior, don't use a global regex as input to the RegExpRecognizer. We should update our docs to reflect that.

Source: [MDN - RegExp.prototype.exec()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec)

>If the match succeeds, the exec() method returns an array and updates properties of the regular expression object. The returned array has the matched text as the first item, and then one item for each capturing parenthesis that matched containing the text that was captured.

>If the match fails, the exec() method returns null.

See also:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec#Finding_successive_matches

>If your regular expression uses the "g" flag, you can use the exec() method multiple times to find successive matches in the same string. When you do so, the search starts at the substring of str specified by the regular expression's lastIndex property (test() will also advance the lastIndex property).

PR #3128, opened 2017-07-13
[Node SDK] Fix, cxt is undefined

abdula:
Fix: ctx is undefined


stevengum97:
@abdula can you transpile the typescript? Thanks.

Issue #1292, opened 2016-09-21
[Feature Request] Docs > How to build Node SDK from source


morsh:
Hi,
I want to edit and offer a pull request to the node.js module, but the development process is unclear to me.
1. It seems that there are missing files (specifically immited ".vscode/*" files that define how to compile the ts files and generate the js code and d.ts files.
2. The root directory of the npm module is unclear (Node/ or Node/core/)
3. Is there anything else I need to know when building the project for local development?

Would really appreciate some help.

Thanks,
Mor

morsh:
Hi @Stevenic, anything new?

palashjhabak:
I would like to know this as well. @morsh, were you able to achieve this?

Stevenic:
We'll add some docs on how to build. Valid criticism.  You'll need VSCode + TypeScript.  Once you install VSCode, the first time you try to build (ctrl+b) it will prompt you to configure your TypeScript task runner. The defaults should work.

Stevenic:
You can also build from the command line by just installing typescript:

```
npm install --g typescript
```

And then running `tsc` from the `src` folder. Output files go to the `build` folder and should be manually copied to `lib`.

morsh:
Hi @Stevenic 
[@palashjhabak]
I've tried above method (after pull).

There are a few issues::
1. I'm getting errors:
`
$ tsc
Session.ts(143,58): error TS2339: Property 'toString' does not exist on type 'never'.
bots/ChatConnector.ts(320,73): error TS2339: Property 'toString' does not exist on type 'never'.
bots/ChatConnector.ts(402,77): error TS2339: Property 'toString' does not exist on type 'never'.
bots/ChatConnector.ts(409,21): error TS2339: Property 'code' does not exist on type 'Error'.
bots/Library.ts(91,22): error TS2405: The left-hand side of a 'for...in' statement must be of type 'string' or 'any'.
bots/Library.ts(92,25): error TS2342: An index expression argument must be of type 'string', 'number', 'symbol', or 'any'.
bots/UniversalBot.ts(525,60): error TS2339: Property 'toString' does not exist on type 'never'.
dialogs/EntityRecognizer.ts(260,29): error TS2339: Property 'toString' does not exist on type 'never'.

dialogs/IntentDialog.ts(261,65): error TS2339: Property 'toString' does not exist on type 'never'.
dialogs/IntentDialog.ts(314,58): error TS2339: Property 'toString' does not exist on type 'never'.
[...]

PR #3138, opened 2017-07-15
[Node SDK] Add custom errors with metadata on ChatConnector

bilby91:
This enables the cleanup of potencial invalid data.

Use case:

When a channel is deleted in Slack there is no event that gets triggered
into the bot. This means that any address associated with the channel
that is cached no longer will be valid. When trying to send a message to
this address, the global handler error will emit an error without
metadata about the action that triggered the error. Using this new type of error, the metadata can
be introspected and be used for taking actions.

```
bot.on('error', (error) => {
    if (error.action == 'postMessage') {
      const address = error.payload.msg.address

      removeFromCache(address)
    }
  })
```

I wasn't able to do in typescript because I can't get the source code to compile. Their is an open issues about that #1292


fcioffi:
Hi guys,

when do you think to merge?

Francesco

PR #3156, opened 2017-07-18
nullpointer exception due to not initializing of ctx

lolo8304:
happens if on slack the conversation is reset (automatically or via /deleteprofile command.
null point exception brings down node


lolo8304:
seems to be the same bug than https://github.com/Microsoft/BotBuilder/pull/3128 but I have fixed it in a different way.
Fixing this bug is important because we were unable to chat on slack in the same channel again even if the bot instance was brought down. /deleteprofile resolved in the same bug.

solution: I had to clone the butbuilder nodemodule to be able to have a successfull chat bot running in slack. check out our bot at: https://github.com/lolo8304/emploji-bot/tree/master/modules/botbuilder

Stevenic:
I took the other fix as we don't need to clone the context in the else path.

PR #3193, opened 2017-07-25
[Node.js SDK] Add next() and missing end() calls to ChatConnector

bilby91:
When using express or restify, in order to get the following handler to
execute, next function needs to be called.

### DIFF
--- a/Node/core/lib/bots/ChatConnector.js
+++ b/Node/core/lib/bots/ChatConnector.js
@@ -38,9 +38,9 @@ var ChatConnector = (function () {
     }
     ChatConnector.prototype.listen = function () {
         var _this = this;
-        return function (req, res) {
+        return function (req, res, next) {
             if (req.body) {
-                _this.verifyBotFramework(req, res);
+                _this.verifyBotFramework(req, res, next);
             }
             else {
                 var requestData = '';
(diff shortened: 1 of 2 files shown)


bilby91:
Any updates on merging this ?

bilby91:
Any update?

JasonSowers:
Will ping someone today from the node side.

bilby91:
@Stevenic Sure, will double check! I doubt that there will be any issues based on conflicts. Will ping soone

bilby91:
@Stevenic We are good to go now! Really nice to see our PRs going in :)

JasonSowers:
@bilby91 we have time scheduled to review the Node PRs Wednesday

harshithkashyap:
@bilby91 Since you are already ending the response with `res.end()`, could you share some info on why you are calling `next()`? 
The `next()` method calls further express middleware functions such as 404 error handler.
#3926 is related to this and we should move the middleware before the `app.post('/api/messages', connector.listen());`.

bilby91:
@harshithkashyap Logging purposes. You want to call next so that a logging handler can capture all the information about the request + response in the logging message.

Make sense ?

PR #3392, opened 2017-08-28
[Node.js SDK] Allow disable batch for serverless operation

ittaibaratz:
Allow disable batch for serverless operation using Apache Openwhisk


ittaibaratz:
Thank you.

PR #3398, opened 2017-08-31
[Node.js SDK] Post process events

ittaibaratz:
For serverless implementation, add an event when batch processing is complete.
Also optimize batch processing to do nothing when there are no messages in the queue, and avoid setTimeout if delay is set to 0.

### DIFF
--- a/Node/core/lib/Session.js
+++ b/Node/core/lib/Session.js
@@ -296,7 +296,7 @@ var Session = (function (_super) {
         this.logger.log(this.dialogStack(), 'Session.endConversation()');
         var ss = this.sessionState;
         ss.callstack = [];
-        this.sendBatch();
+        this.sendBatch(_this._onBatchComplete);
         return this;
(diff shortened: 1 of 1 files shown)


Node/core/lib/Session.js
```diff
@@ -407,6 +407,11 @@ var Session = (function (_super) {
     Session.prototype.sendBatch = function (done) {
         var _this = this;
         this.logger.log(this.dialogStack(), 'Session.sendBatch() sending ' + this.batch.length + ' message(s)');
+        if (this.batch.length == 0) {
+        		// No messages to send - Nothing to do.
+        		return;
```
stevengum97: Please correct the spacing on these lines 411 and 412.

ittaibaratz: I will do a new PR with both #3392 and #3398 including both Typescript and Javascript.

PR #3478, opened 2017-09-18
Update bot-framework-user-agent.md

johnd-ms:
Redirect to published point in our public docs.


en-us/bot-framework-user-agent.md
```diff
@@ -1,27 +1,6 @@
 ---
-layout: page
-title: Bot Framework User-Agent
-permalink: /en-us/bot-framework-user-agent/
+layout: 'redirect'
+permalink: /en-us/
```
RobStand: Permalink should be the same as the original one.

Issue #3508, opened 2017-09-24
bot.loadSession no longer behaving properly in 3.10.1


techno246:
## Bot Info 
* SDK Platform: Node.js
* SDK Version: 3.10.1
* Active Channels: Facebook
* Deployment Environment: Elastic Beanstalk

## Issue Description
`bot.loadSession(address, done)` is breaking on the latest release. When the function completes, the handler (done) is called twice - one with the session object, and one where it is set to null. Calling `session.send()` doesn't seem to work even in the instance where the session object exists. I have verified that the address exists and is valid.

Looking at the latest commits, it appears that loadSession has been refactored to call `loadSessionWithOptionalDispatch(address, false, done)`, to incorporate the feature of optional dispatching.  However, shouldn't it call `loadSessionWithOptionalDispatch(address, true, done)` instead, since loadSession used to dispatch by default?

Reverting to 3.9.1 fixed the issue.

dvorakjan:
exactly the same for me

stevengum97:
I was able to reproduce the problem and have created a PR that switches the 2nd parameter to true, which has `UniversalBot.loadSession()` returning to its original behavior. 

PR that changed behavior: #3396

dan-cohn:
I'm having a similar problem, although I don't see the callback being called twice. In my case, it gets called with the following error -- `TypeError: Cannot read property 'callstack' of undefined`

This was working fine in 3.9.1.

nwhitmont:
resolved in botbuilder@3.10.2 out now

PR #3525, opened 2017-09-26
[Node.js SDK] fix bot.loadSession to use dispatch by default

stevengum97:
Addresses #3508, which arose from #3396


nwhitmont:
closed #3508

PR #3618, opened 2017-10-15
[Node.js SDK] Added support for 'MultiTypes' prompt

eladiw:
The MultiTypes prompt allow the bot creator to use a prompt which expects any combination of the supported input types (i.e. attachment, date, number or text).
It makes scenarios in which the input type is dynamic much more simple (no need to display the 'choice' prompt prior to the actual prompt)

Experience example:
screen shot 2017-10-15 at 13 57 12

Usage example (code):
screen shot 2017-10-15 at 13 59 00


stevengum97:
@eladiw, Thank you for the contribution but I have some reservations about this PR.

Personally I think that one or all of the accepted inputs should be enabled by default ([line 26][L26]), but that's because I'd rather users don't have to specify something a Prompt implicitly says it does. I would expect a "multiType" prompt to already _accept_ multiple inputs, and not need me to turn it on.

I'm also worried about the logic for some of your if statements. For example, on [line 70][L70] of PromptMultiTypes.ts, this logic is skipped if there are commas in their context.message.text or if they use numerals (e.g. `"two"`) as both of these cases return `NaN` when plugged into `Number()`. 

For [line 64][L64] (for the attachment logic) there isn't any `contentType` check, which is a feature of PromptAttachment. If a chatbot is expecting a picture but receives a music file instead, this should cause the bot to reprompt the user.

Additionally, not every user's situation will need the same priority chain: first attachments, then datetime, then numbers, then text. 

I think this is a great example/case for using a non-standard prompt, but I'm not sure this is the direction we want to head in. 

@Stevenic what are your thoughts on this PR?

  [L70]: https://github.com/Microsoft/BotBuilder/pull/3618/files#diff-100a7d373253b2c63e1ef4cdb198c3ddR70
  [L64]: https://github.com/Microsoft/BotBuilder/pull/3618/files#diff-100a7d373253b2c63e1ef4cdb198c3ddR64
  [L26]: https://github.com/Microsoft/BotBuilder/pull/3618/files#diff-96aa5004cae82b53ba928d008dba5e0aR26

eladiw:
Hi @stevengum97, Thanks for the quick review.
I agree about the fact that it makes more sense to enable all input types in advance. (will fix)

I will fix the issue on line 70, I am aiming at either using an existing package which will probably deal with situations such as 'Two' and commas quite well, or I would just reuse the code from the number prompt...

regarding the contentType, agreed that acceptable content types should also be passed and verified.

Regarding the priority chain. The decision was from the most restrictive to the least, I guess this also can be configurable, but I believe that the current order should be ideal in most cases (in the next iteration I will leave it as-is, unless you have a strong opinion against.)

Lastly, Considering the fact that adding a dynamic prompt makes some flows much more intuitive and reduce the number of prompts (- choice prompt), **_IMO_**, This will probably be a good addition to the SDK.

Stevenic:
This is an interesting idea Elad... For us to consider this as part of the SDK it would need to use our built-in entity recognizers. I'd actually recommend that for now you package this up as a separate library that developers can add to their project and then give us some time to reflect over supporting this directly within the SDK. Like I said... It's a very interesting idea. Feel free to ping me internally over email (@stevenic) if you'd like more context.

Steve

eladiw:
Thanks @aribornstein for your help in the PR

eladiw:
Hi @stevengum97, @Stevenic,
I have incorporated your suggestions into the PR:
a) I am using the existing recognizers
b) the priority is determined based on the scores, text is chosen if all other options are not working
c) default now allows all input types

Let me know if you have any concerns / questions.

screen shot 2017-10-17 at 13 39 48

Stevenic:
Hi Elad... We discussed this internally and while its an interesting idea we're not sure how much value it actually delivers given that you're usually expecting a very specific answer to a prompt. If you want to package this up as a separate extension to the SDK I think that would be great :)

PR #3626, opened 2017-10-17
Add null check to make sure prompt dialog on top of stack is available

raghavan:
The module to forward message from bot to the user verifies if there is any prompt dialog to check if the bot is expecting any input. This prompt dialog may be null if the resume after function is static and it requires a null check to avoid null references.


EricDahlvang:
@raghavan Thanks for submitting a fix for this.

PR #3690, opened 2017-10-30
[.NET Example] Update PizzaOrderDialog.cs

danielsemedo:
Comments of how to get a key in LUIS services.


danielsemedo:
Removed the keys and commented how to get a new key at Luis Services website.

JasonSowers:
We have provided the app id and key so that people can plug and play this example.  Thank you for your contribution, but we are not going to accept this at this time.  
  
If you would like to update your pull request with your comments and leaving the keys in the sample we can accept this.

Issue #3516, opened 2017-09-25
[Node.js SDK] EntityRecognizer findBestMatch returns score > 1


ubreddy:
## Bot Info 
* SDK Platform: Node.js
* SDK Version: 3.9.1
* Active Channels: web
* Deployment Environment: Azure App Service

## Issue Description
`botbuilder.EntityRecognizer(array, value)` returns score > 1 
Perfect matches return 1
some  near perfect matches return > 1

## Code Example

```js
syns = ["email link", "link", "email", "Through a link to salesforce in an email message", "3", "three", "c"]
inputOption = "Through a link to salesforce in an email message" 
var match = builder.EntityRecognizer.findBestMatch(syns, inputOption, 0.65);
// match.score = 1.2
```

## Expected Behavior
a score <= 1 as defined in specs

stevengum97:
@ubreddy, I was able to reproduce this, I'm looking into a fix now.

ubreddy:
any expected time to fix this?

stevengum97:
Sorry for the delay, we're working on a fix for the underlying bug that's causing this behavior.

stevengum97:
@ubreddy We're planning a new release for next week which will contain the fix I wrote. That said, we also recommend moving towards using the `PromptRecognizers `class which does feature improved recognition, and has been used in the new Prompt system since **3.8**. (**3.7.x** and earlier used the `EntityRecognizer` class for its recognition)

Closing this issue, if the fix or the `PromptRecognizers` class doesn't provide what you're looking for, feel free to reopen this issue and mention me.

PR #3702, opened 2017-11-01
[Node.js SDK] EntityRecognizer.findAllMatches fix to prevent scores > 1.0 and adjust score generation

stevengum97:
Addresses underlying cause of #3516.

### DIFF
--- a/Node/core/lib/dialogs/EntityRecognizer.js
+++ b/Node/core/lib/dialogs/EntityRecognizer.js
@@ -204,6 +204,7 @@ var EntityRecognizer = (function () {
                     }
                 }
                 score = tokenScore / value.length;
+                score = score > 1 ? 1 : score;
             }
             if (score >= threshold) {
                 matches.push({ index: index, entity: choice, score: score });
--- a/Node/core/lib/dialogs/EntityRecognizer.js
+++ b/Node/core/lib/dialogs/EntityRecognizer.js
@@ -176,13 +176,34 @@ var EntityRecognizer = (function () {
                 score = Math.min(0.5 + (value.length / utterance.length), 0.9);
             }
             else {
-                var matched = '';
+                var matched = {};
                 tokens.forEach(function (token) {
                     if (value.indexOf(token) >= 0) {
-                        matched += token;
+                        if (!matched[token]) {
+                            matched[token] = 1;
+                        }
                     }
                 });
-                score = matched.length / value.length;
+                var tokenizedValue = value.split(' ');
+                var tokenScore = 0;
+                for (var token in matched) {
+                    tokenizedValue.forEach(function (val) {
+                        if (val.indexOf(token) >= 0 && token.length <= val.length / 2) {
+                            matched[token]--;
+                        }
+                        else if (val.indexOf(token) == -1) {
+                        }
+                        else {
+                            matched[token]++;
+                        }
+                    });
+                }
+                for (var token in matched) {
+                    if (matched[token] > 0) {
+                        tokenScore += token.length;
+                    }
+                }
+                score = tokenScore / value.length;
             }
             if (score >= threshold) {
                 matches.push({ index: index, entity: choice, score: score });
--- a/Node/core/src/dialogs/EntityRecognizer.ts
+++ b/Node/core/src/dialogs/EntityRecognizer.ts
@@ -266,6 +266,7 @@ export class EntityRecognizer {
                     }
                 }
                 score = tokenScore / value.length;
+                score = score > 1 ? 1 : score;
             }
             if (score >= threshold) {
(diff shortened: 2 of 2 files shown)


Node/core/lib/dialogs/EntityRecognizer.js
```diff
@@ -176,13 +176,34 @@ var EntityRecognizer = (function () {
...
+                var matched = {};
                 tokens.forEach(function (token) {
                     if (value.indexOf(token) >= 0) {
-                        matched += token;
+                        if (!matched[token]) {
+                            matched[token] = 1;
+                        }
                     }
                 });
-                score = matched.length / value.length;
```
stevengum97: The problem from calculating the score this way is that on [line 181][L181] it will add words that shouldn't be added to `matched`. 

In #3516 `EntityRecognizer.findAllMatches()` added `"a"` and `"in"` to `matched`, because these two words can be found in `"email"` and `"link"`.  Thus `matched` would equal `"alinkinemail"`. The entity value the user was looking for was `"email link"` and when `matched.length` is divided by `value.length`, the returned score is 1.2.


  [L181]: https://github.com/Microsoft/BotBuilder/pull/3702/files#diff-a5cbf2306473b89ce8f6eb849e7f09c5L181

Node/core/lib/dialogs/EntityRecognizer.js
```diff
@@ -176,13 +176,34 @@ var EntityRecognizer = (function () {
                 score = Math.min(0.5 + (value.length / utterance.length), 0.9);
             }
             else {
-                var matched = '';
+                var matched = {};
                 tokens.forEach(function (token) {
                     if (value.indexOf(token) >= 0) {
-                        matched += token;
+                        if (!matched[token]) {
+                            matched[token] = 1;
```
stevengum97: On [line 179][L179] we change `matched ` to an object and on [line 183][L183] we add a key-value pair to `matched` if the token was not already found in `matched`.


  [L179]: https://github.com/Microsoft/BotBuilder/pull/3702/files#diff-a5cbf2306473b89ce8f6eb849e7f09c5R179
  [L183]: https://github.com/Microsoft/BotBuilder/pull/3702/files#diff-a5cbf2306473b89ce8f6eb849e7f09c5R183

Node/core/lib/dialogs/EntityRecognizer.js
```diff
@@ -176,13 +176,35 @@ var EntityRecognizer = (function () {
...
+                        }
+                    });
+                }
+                for (var token in matched) {
+                    if (matched[token] > 0) {
+                        tokenScore += token.length;
+                    }
+                }
+                score = tokenScore / value.length;
+                score = score > 1 ? 1 : score;
```
stevengum97: For [line 189][L189] through [line 200][L200] we evaluate if the `token` from `matched` exists in and is less than half of the current token (`val`) from the earlier variable `value`. If these are both true, we're assuming it's not relevant to the current value, and we decrement `matched[token]`. If the `token` from `matched` doesn't exist in the current `val`, do nothing. Passing these two cases, the [current `token`][L197] must be larger than half the current `val` and therefore is relevant to the current `val` (see [Stemming and lemmatization][Stanford] from the Stanford NLP Group).

On [line 201][L201] through [line 205][L205] we evaluate which `tokens` are still relevant to the current `value` (relevant meaning they have a score greater than 0). If relevant we'll add the length of the token to `tokenScore`. We then calculate the score in the same way as the previous method (dividing the length of all of the matching tokens by the length of the current `value`). 

As an added precaution, there's a ternary operator for `score` on [line 207][L207] that evaluates if the `score > 1` in which case it changes score to 1.

  [L189]: https://github.com/Microsoft/BotBuilder/pull/3702/files#diff-a5cbf2306473b89ce8f6eb849e7f09c5R189
  [L197]: https://github.com/Microsoft/BotBuilder/pull/3702/files#diff-a5cbf2306473b89ce8f6eb849e7f09c5R197
  [L200]: https://github.com/Microsoft/BotBuilder/pull/3702/files#diff-a5cbf2306473b89ce8f6eb849e7f09c5R200
  [Stanford]: https://nlp.stanford.edu/IR-book/html/htmledition/stemming-and-lemmatization-1.html
  [L201]: https://github.com/Microsoft/BotBuilder/pull/3702/files#diff-a5cbf2306473b89ce8f6eb849e7f09c5R201
  [L205]: https://github.com/Microsoft/BotBuilder/pull/3702/files#diff-a5cbf2306473b89ce8f6eb849e7f09c5R205
  [L207]: https://github.com/Microsoft/BotBuilder/pull/3702/files#diff-a5cbf2306473b89ce8f6eb849e7f09c5R207

Issue #3845, opened 2017-11-29
[Feature Request] Add the IgnoreField attribute to the FormFlow


flash2048:
## Issue Description
For large classes, it would be convenient to have an attribute that allows you to remove the marked field from processing.

## Code Example
`[IgnoreField]`
`public int Id;`

`[IgnoreField]`
 `public string PasswordHash;`

## Actual Results
Convenient and quick exclusion of extra fields from the class when using FormFlow

stevengum97:
@flash2048, since you've already filed a PR and it's been assigned a reviewer, I'm closing this issue for now. Thanks for the contribution!

PR #3846, opened 2017-11-29
[.NET SDK] Adding the IgnoreField attribute to the FormFlow #3845

flash2048:
IgnoreField attribute for convenient and quick exclusion of extra fields from the class when using FormFlow


JasonSowers:
Thank you for your contribution

PR #3849, opened 2017-11-29
[Node.js SDK] Add batchComplete event to support serverless deployments

ittaibaratz:
Add batchComplete event to support serverless deployments where we need to freeze a container after messages have been sent out.

Also avoid starting a batch when there are no messages in the queue.

@stevengum97 - This is a redo of #3398 and #3392 in TypeScript.


Stevenic:
This would break `session.save()` and is a really big change.  I'm not sure how comfortable I am taking any change (even if it seems minor) to the way batching works. It's taken us a long time to get this debugged and stable.

EricDahlvang:

PR #3879, opened 2017-12-06
add next fix to ChatConnector.dispatch()

stevengum97:
Add fix to `ChatConnector.dispatch()` regarding `next`.
- If `next` is falsey, `next` will default to `function() { };`
- This covers cases where `next` is not provided, e.g. when not using restify to serve a bot.

### DIFF
--- a/Node/core/src/bots/ChatConnector.ts
+++ b/Node/core/src/bots/ChatConnector.ts
@@ -549,6 +549,7 @@ export class ChatConnector implements IConnector, IBotStorage {
 
     private dispatch(msg: IMessage, res: IWebResponse, next: Function) {
         // Dispatch message/activity
+        next = next || function () {};
         try {
             this.prepIncomingMessage(msg);
             logger.info(msg, 'ChatConnector: message received.');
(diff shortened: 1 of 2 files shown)


Node/core/src/bots/ChatConnector.ts
```diff
@@ -549,6 +549,7 @@ export class ChatConnector implements IConnector, IBotStorage {
 
     private dispatch(msg: IMessage, res: IWebResponse, next: Function) {
```
santiagodoldan: `private dispatch(msg: IMessage, res: IWebResponse, next = () => void) {` not sure if it does  the same

stevengum97: I haven't tried that but it is an interesting item. I'm not sure an arrow function would work in this context, it would be interesting to see what `tsc` compiles it to.

santiagodoldan: Yep, I was testing that because I wasn't sure either. Looks like you'd need to change the other signatures from `next: Function` to `next: () => void` and here use `next = () => {}`

It generates this `if (next === void 0) { next = function () { }; }`

stevengum97:
Closing PR since [Commit eca1810](https://github.com/Microsoft/BotBuilder/commit/eca181033aa688b3906ef249cf1315420a291176) addresses the issue this PR was addressing.

Node/core/src/bots/ChatConnector.ts
stevengum97: Thanks for the feedback and info! As indicated I closed this PR due to Steve adding some changes to a different method which addresses the same problem.

Issue #78, opened 2016-04-08
Last Seperator && Seperator not honored


taylorgibb:
```
[Prompt("Is that a {||}?", "I am not sure if that is a {||}?", "Help me out a little bit. Is that a {||}?", AllowDefault = BoolDefault.False, ChoiceFormat = "{1}", ChoiceStyle = ChoiceStyleOptions.Inline, Feedback = FeedbackOptions.Never, FieldCase = CaseNormalization.Default, LastSeparator = " or ", Separator = ",")]
```

Outputs: I am not sure if that is a (Movie, Series, Song)?

I also tried: 

```
            builder.Configuration.DefaultPrompt.Separator = ", ";
            builder.Configuration.DefaultPrompt.LastSeparator = " or ";
            builder.Configuration.DefaultPrompt.ChoiceFormat = "{1}";
            builder.Configuration.DefaultPrompt.ChoiceStyle = ChoiceStyleOptions.Inline;
            builder.Configuration.DefaultPrompt.ValueCase = CaseNormalization.Lower;
```

Expected result: I am not sure is that is a movie, series or song?

Dont know if am doing something wrong, if not, and someone can confirm its a bug and point me in the right direction i can take a look at it.

chrimc62:
Separator and LastSeparator are only used when generating lists of values from your underlying state.  In this case you are looking for them to apply to the listing of possible choices. I'll have to think about the right way to improve this.  I don't think using the settings for both makes sense since {Toppings} should be "Pepperoni, Onions and Cheese" whereas {||} might be "(Thin, thick or filled)". The right thing to do is probably to add LastChoiceSeparator and ChoiceSeparator to the prompt and use those.  I'll make that change.

taylorgibb:
Agreed. I wasn't entirely sure, maybe we can update the documentation around that as well once we have these new attributes.

taylorgibb:
Also, the ability to control casing, and whether to enclose it in parenthesis would be good as well.

chrimc62:
The documentation on Separator and LastSeparator does indicate it only applies to {[]} expressions, but the table in the description indicated they applied to both--I've fixed that.  I've also implemented ChoiceSeparator, ChoiceLastSeparator, ChoiceCase and ChoiceParens.  It is in review and should be checked into the develop branch soon to go out with our next release.  Thanks for the feedback!

PR #3890, opened 2017-12-07
[.NET SDK] Add support for BingSpellCheckSubscriptionKey to LuisOptions, LuisSevice, LuisModel

stevengum97:
Per LUIS Runtime Swagger File, [Line #78](https://github.com/Azure/azure-rest-api-specs/blob/current/specification/cognitiveservices/data-plane/LUIS/Runtime/v2.0/LUIS-Runtime.json#L78)

```js
{
  "name": "bing-spell-check-subscription-key",
  "in": "query",
  "description": "The subscription key to use when enabling bing spell check",
  "type": "string"
}
```


JasonSowers:
Seems straightforward

Issue #3917, opened 2017-12-13
[Code Examples] C# Samples are not building, error "cannot restore Nuget package"?


gunnicus:
## Bot Info 
* SDK Platform: .NET

## Issue Description
Samples are not building

## Reproduction Steps
1. Dowanload BotBuilder-master from github, unzip, Open Microsoft.Bot.Builder.Samples.sln with VS2017 community
2. Press F5
3. See errors 
4. Manually add reference, still get same error

## Expected Behavior
Build success.

## Actual Results
Severity	Code	Description	Project	File	Line	Suppression State
Error		This project references NuGet package(s) that are missing on this computer. Use NuGet Package Restore to download them.  For more information, see http://go.microsoft.com/fwlink/?LinkID=322105. The missing file is ..\..\packages\Microsoft.Net.Compilers.1.0.0\build\Microsoft.Net.Compilers.props.	Microsoft.Bot.Sample.AlarmBot	D:\ProjectCode\BotBuilder-master\CSharp\Samples\AlarmBot\Microsoft.Bot.Sample.AlarmBot.csproj	243	

![image](https://user-images.githubusercontent.com/4023477/33919984-a0611f2e-e00f-11e7-8d20-429360283a42.png)

![image](https://user-images.githubusercontent.com/4023477/33919994-ad8a1638-e00f-11e7-88b2-891120b8ac8c.png)

FranciscoPonceGomez:
Hi @gunnicus,

There seems to be a problem related to the directory where the nugget packages are being restored. The projects are looking for the packages folder two levels up from them but the packages are being restores in the immediate superior level. 
We are working on a fix but for now you can right click on the project and restore the packages, and then move or copy the packages folder to an upper level so the project can find the restored packages in ..\packages instead of ..\..\packages:

![image](https://user-images.githubusercontent.com/16775093/33962484-9445a3b2-e006-11e7-851f-2a4faa5b1b94.png)
![image](https://user-images.githubusercontent.com/16775093/33962498-a2535fee-e006-11e7-89b9-881bdfaaa66c.png)

Please, let me know if that fixes the issue.
Francisco

gunnicus:
Thanks for your reply,

Tired copy packages folder to different location, didn't work. Tried remove the reference and manually add back, didn't work neither. 

1. 
Remove the reference
![image](https://user-images.githubusercontent.com/4023477/33966008-12189a14-e0b2-11e7-89b4-6c00f47dce68.png)

2.
[...]

PR #3929, opened 2017-12-15
Adds NuGet.config for samples

EricDahlvang:
This will ensure packages restore to the correct directory, and update package references for AnnotatedSandwichBot.


EricDahlvang:
fix for https://github.com/Microsoft/BotBuilder/issues/3917

PR #4039, opened 2018-01-12
[.NET SDK] ReplayTranscript constructor ignores the "header" argument

artemious7:
ReplayTranscript constructor ignores the "header" argument, because it does not assign the argument to the corresponding field, and it uses the default "header" function.


stevengum97:
Looks good, thanks for the contribution!

PR #4120, opened 2018-01-30
Add startReplyChain to UniversalBot and IConnector

bilby91:
At the moment the following connectors implement the startReplyChain:

- botbuilder-teams
  (https://github.com/OfficeDev/BotBuilder-MicrosoftTeams)
- botbuilder-slack (https://github.com/suttna/botbuilder-slack)

On both cases, the current solution at the connector level breaks the
usage of middlewares. In both cases, the "send" middleware is not called
when sending a message using the startReplyChain api.

We want to standarize the startReplyChain since a lot of clients support
it and is very useful operation. The idea would be to get this merged so
that both previously mentioned libraries can update the implementation
to match the new signature.

### DIFF
--- a/Node/core/lib/botbuilder.d.ts
+++ b/Node/core/lib/botbuilder.d.ts
@@ -1356,6 +1356,15 @@ export interface IConnector {
     send(messages: IMessage[], callback: (err: Error, addresses?: IAddress[]) => void): void;
 
     /**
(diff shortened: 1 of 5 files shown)


bilby91:
@JasonSowers want to know your thoughts about this.

Thanks!

EricDahlvang:

PR #4121, opened 2018-01-30
[Node.js SDK] Add onBeforeBegin to WaterfallDialog

santiagodoldan:
Here just a usage sample

```javascript
  import { Prompts, WaterfallDialog } from "botbuilder"

  const beforeBegin = async (session: Session, args: any, next: (args: any) => void) => {
    console.info("beforeBegin")

    next({ ...args, newArg: "newArg" })
  }

  const beforeStep = async (session: Session, step: number, args: any, next: (step: number, args: any) => void ) => {
    console.info("beforeStep")

    next(step, { ...args, anotherArg: "anotherArg" })
  }

  return new WaterfallDialog([
    async (session: Session) => {
      Prompts.confirm(session, "Do you want to continue?")
    }, async (session: Session) => {
      session.say("Ok")
    },
  ]).onBeforeBegin(beforeBegin).onBeforeStep(beforeStep)
```

### DIFF
--- a/Node/core/lib/botbuilder.d.ts
+++ b/Node/core/lib/botbuilder.d.ts
@@ -3930,6 +3930,14 @@ export class WaterfallDialog extends Dialog {
     replyReceived(session: Session): void;
 
     /**
+     * Registers a handler that will be called before the first waterfall step. This handler lets a
+     * developer skip the entire dialog and process the args being passed to the first step of it.
+     *
+     * @param handler Function to invoke before the first waterfall step.
+     */
+    onBeforeBegin(handler: (session: Session, args: any, next: (args: any) => void) => void): WaterfallDialog;
+
+    /**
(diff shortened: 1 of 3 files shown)


santiagodoldan:
@stevengum97 any thoughts?

JasonSowers:
Could you please provide use cases, and test cases you used to test this.

Stevenic:
Why not just create a new WaterfallDialog class that provides this?  I'm not completely sure of the use case here. Are you trying to add logging?

bilby91:
@Stevenic You mean a subclass of WaterfallDialog ?

We are using this to load information from the database on every interaction with that WaterfallDialog. So, basically on every step execution we need to use the same data from the database. 

Make more sense ?

EricDahlvang:

Issue #4128, opened 2018-01-31
DictionaryBotData doesn't work because of deserialisation 


gareththackeray:
## Bot Info
* SDK Platform: .NET
* SDK Version: latest
* Active Channels: any
* Deployment Environment: any

## Issue Description
In the bot builder code there is an `IBotData` implementation called `DictionaryBotData`.  This implementation doesn't work when the `IBotDataBag.Data` (which is an `object`) gets deserialised from the store it gets deserialised to a `JObject`.

## Code Example
       
            Conversation.UpdateContainer(builder =>
            {
                builder.RegisterKeyedType<DictionaryBotData, IBotData>()
                    .AsSelf()
                    .InstancePerLifetimeScope();

                builder.RegisterAdapterChain<IBotData>
                    (
                        typeof(DictionaryBotData),
                        typeof(DialogTaskManagerBotDataLoader)
                    )
                    .InstancePerLifetimeScope();

                builder.Register(c => new DialogTaskManager(DialogModule.BlobKey,
                    c.Resolve<DictionaryBotData>(),
                    c.Resolve<IStackStoreFactory<DialogTask>>(),
                    c.Resolve<Func<IDialogStack, CancellationToken, IDialogContext>>(),
                    c.Resolve<IEventProducer<IActivity>>())).AsSelf()
                    .As<IDialogTaskManager>()
                    .As<IDialogTasks>()
                    .InstancePerLifetimeScope(); ;
            });

## Reproduction Steps
1. Send 2 messages to a bot with the above overrides.

## Expected Behavior
Bot should work as normal.

## Actual Results
On the 2nd message there will be a cast exception from `BotDataBase<Dictionary<string, object>>.LoadAsync()`.

gareththackeray:
Created PR https://github.com/Microsoft/BotBuilder/pull/4129

FranciscoPonceGomez:
Hi gareththackeray,

Thanks for bringing up this issue and providing a fix. I will review it as soon as possible.

Regards,
Francisco

FranciscoPonceGomez:
Can I ask you if there is any reason why you can't use JObjectBotData class instead of DictionaryBotData for your solution?

https://github.com/Microsoft/BotBuilder/blob/db2b8f860a3d8f7744a378930f93c4b0baa97978/CSharp/Library/Microsoft.Bot.Builder/ConnectorEx/BotData.cs#L633

gareththackeray:
[...]

PR #4129, opened 2018-01-31
Allow DictionaryBotData to work

gareththackeray:
Because of serialising a JSON object to a .net object creates a JObject, we cannot expect botData.Data to be T.  Hence deal with it being a JObject.

Note we need to allow for `botData.Data` being either `T` or `JObject` because in the case that it is returned from the `CachingBotDataStore` it will be a `T` while if it's been loaded from a persistent store where it was serialised it will be a `JObject`.  

This is a fix for https://github.com/Microsoft/BotBuilder/issues/4128


FranciscoPonceGomez:
Hi gareththackeray,

I am going to close this PR. Please, comment in the Issue thread if you have any further problems.

Regards,
Francisco

PR #4158, opened 2018-02-07
Similar to activity type "Typing" we want to provide antoher activity…

amitkumarshukla:
… type which is "Thinking". This mainly suits the scenario where the user and the bot both are doing communication using audio (both of them are speaking". There is no typing involved.


JasonSowers:
Please resolve conflicts and provide any test cases you used to test

Issue #3261, opened 2017-08-04
[Feature Request] Add missing UniversalBot constructor signatures in botbuilder.d.ts


jorenl:
## System Information
* SDK Language: Node.js
* SDK Version: latest

## Issue Description
While in [the implementation](https://github.com/Microsoft/BotBuilder/blob/master/Node/core/src/bots/UniversalBot.ts#L94), UniversalBot constructor signatures exist allowing you to pass in a settings object (IUniversalBotSettings), for example to use a different IBotStorage than the one provided by the ChatConnector, the .d.ts file does not seem to include them, so my typescript code won't compile out of the box.

https://github.com/Microsoft/BotBuilder/blob/master/Node/core/lib/botbuilder.d.ts#L3859

Is this intentional?
If not, and the .d.ts is manually maintained and not auto-generated, I'd be happy to submit a PR.

## Code Example
```
import * as bb from 'botbuilder';

let storage: bb.IBotStorage = new MyBotStorage();
let bot = new bb.UniversalBot(connector, {storage: storage});
```
Produces:
```
severity: 'Error'
message: 'Argument of type '{ storage: IBotStorage; }' is not assignable to parameter of type 'IDialogWaterfallStep | IDialogWaterfallStep[]'.
  Object literal may only specify known properties, and 'storage' does not exist in type 'IDialogWaterfallStep | IDialogWaterfallStep[]'.'
at: '305,47'
source: 'ts'
```

nwhitmont:
Thanks for reporting @jorenl - Feel free to submit a PR if you have the fix.

stevengum97:
Closing issue as #4182 has been merged.

PR #4182, opened 2018-02-09
[Node.js SDK] Added missing constructors for UniversalBot (#3261)

Xzya:
Fix for #3261


JasonSowers:
Thank you for your controbution

PR #4221, opened 2018-02-23
Fix perf issues with HttpClient in SDK

tomlm:
* replace instance HttpClient with singleton in Auth path
* Change RefreshToken to have single Task so only 1 request happens even when under load
* change ConnectorClient and StateClient to use singleton httpclient 
* Bump version to 3.14.0

### DIFF
--- a/CSharp/Library/Microsoft.Bot.Connector.Shared/ActivityEx.cs
+++ b/CSharp/Library/Microsoft.Bot.Connector.Shared/ActivityEx.cs
@@ -40,12 +40,8 @@ namespace Microsoft.Bot.Connector
 
         partial void CustomInit()
         {
-            MembersAdded = MembersAdded ?? new List<ChannelAccount>();
-            MembersRemoved = MembersRemoved ?? new List<ChannelAccount>();
             Attachments = Attachments ?? new List<Attachment>();
             Entities = Entities ?? new List<Entity>();
-            ReactionsAdded = ReactionsAdded ?? new List<MessageReaction>();
-            ReactionsRemoved = ReactionsRemoved ?? new List<MessageReaction>();
         }
 
         /// <summary>
@@ -90,7 +86,14 @@ namespace Microsoft.Bot.Connector
         /// <summary>
         /// Create an instance of the Activity class with IConversationUpdateActivity masking
         /// </summary>
-        public static IConversationUpdateActivity CreateConversationUpdateActivity() { return new Activity(ActivityTypes.ConversationUpdate); }
+        public static IConversationUpdateActivity CreateConversationUpdateActivity()
+        {
+            return new Activity(ActivityTypes.ConversationUpdate)
+            {
+                MembersAdded = new List<ChannelAccount>(),
+                MembersRemoved = new List<ChannelAccount>()
+            };
+        }
 
         /// <summary>
         /// Create an instance of the Activity class with ITypingActivity masking
--- a/CSharp/Library/Microsoft.Bot.Connector.Shared/ConnectorClientEx.cs
+++ b/CSharp/Library/Microsoft.Bot.Connector.Shared/ConnectorClientEx.cs
@@ -31,7 +31,7 @@ namespace Microsoft.Bot.Connector
         /// <param name="addJwtTokenRefresher">True, if JwtTokenRefresher should be included; False otherwise.</param>
         /// <param name="handlers">Optional. The delegating handlers to add to the http client pipeline.</param>
         public ConnectorClient(Uri baseUri, MicrosoftAppCredentials credentials, bool addJwtTokenRefresher = true, params DelegatingHandler[] handlers)
-            : this(baseUri, addJwtTokenRefresher ? AddJwtTokenRefresher(handlers, credentials): handlers)
+            : this(baseUri, addJwtTokenRefresher ? AddJwtTokenRefresher(handlers, credentials) : handlers)
         {
             this.Credentials = credentials;
         }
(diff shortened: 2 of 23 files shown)


CSharp/Library/Microsoft.Bot.Connector.Shared/MicrosoftAppCredentials.cs
```diff
@@ -244,33 +263,32 @@ private OAuthException(SerializationInfo info, StreamingContext context)
...
-                        return oauthResponse;
-                    }
-                    catch (Exception error)
-                    {
-                        throw new OAuthException(body ?? response.ReasonPhrase, error);
-                    }
+                    response.EnsureSuccessStatusCode();
+                    body = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
+                    var oauthResponse = JsonConvert.DeserializeObject<OAuthResponse>(body);
+                    oauthResponse.expiration_time = DateTime.UtcNow.AddSeconds(oauthResponse.expires_in).Subtract(TimeSpan.FromSeconds(60));
```
dandriscoll: This pattern can really confuse things. Instead of storing the actual expiration, this code recalculates what the refresh time should be, and then stores that as if it were the expiration.

I think it should just store the raw expiration and if we're going to do any calculation, we can do when we read the value later.

CSharp/Library/Microsoft.Bot.Connector.Shared/ActivityEx.cs
```diff
@@ -90,7 +86,14 @@ public Activity CreateReply(string text = null, string locale = null)
         /// <summary>
         /// Create an instance of the Activity class with IConversationUpdateActivity masking
         /// </summary>
-        public static IConversationUpdateActivity CreateConversationUpdateActivity() { return new Activity(ActivityTypes.ConversationUpdate); }
+        public static IConversationUpdateActivity CreateConversationUpdateActivity()
+        {
+            return new Activity(ActivityTypes.ConversationUpdate)
+            {
+                MembersAdded = new List<ChannelAccount>(),
+                MembersRemoved = new List<ChannelAccount>()
```
dandriscoll: Do you want this in this change?

CSharp/Library/Microsoft.Bot.Connector.Shared/ConnectorClientEx.cs
```diff
@@ -49,25 +49,44 @@ public ConnectorClient(Uri baseUri, MicrosoftAppCredentials credentials, HttpCli
...
-            HttpClient.DefaultRequestHeaders.ExpectContinue = false;
+            if (g_httpClient == null)
+            {
+                lock (syncObj)
+                {
+                    if (g_httpClient == null)
+                    {
+                        g_httpClient = new HttpClient();
+                        g_httpClient.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue("Microsoft-BotFramework", "3.1"));
+                        g_httpClient.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue($"(BotBuilder .Net/{GetClientVersion(this)})"));
```
dandriscoll: If this builds one long User-Agent string ("Microsoft-BotFramework/3.1 (BotBuilder .Net/1.2.3.4)" then awesome. Does it?

CSharp/Library/Microsoft.Bot.Connector.Shared/MicrosoftAppCredentials.cs
```diff
@@ -161,40 +161,59 @@ public static bool IsTrustedServiceUrl(string serviceUrl)
...
+                    }
+                }
             }
-            catch (OAuthException)
+
+            // if we have new token we need to pull it out again
+            if (newAuthTokenTask)
             {
-                throw;
+                oAuthToken = await oAuthTokenTask.ConfigureAwait(false);
```
dandriscoll: In the event that oauthTokenTask failed, this await will throw the failure exception for every thread that hits this line.

CSharp/Library/Microsoft.Bot.Connector.Shared/ConnectorClientEx.cs
```diff
@@ -49,25 +49,44 @@ public ConnectorClient(Uri baseUri, MicrosoftAppCredentials credentials, HttpCli
...
             var handlers = new List<DelegatingHandler>(srcHandlers);
             handlers.Add(new JwtTokenRefresher(credentials));
             return handlers.ToArray();
         }
-        
-        // client defaults to sending the expect: continue header, which isn't very efficient, 
+
+        private HttpClient instanceClient;
+        protected static HttpClient g_httpClient = null;
+        protected static object syncObj = new object();
```
dandriscoll: g_syncObj ?

cleemullins: Could we use one naming standard? :) 

_instanceClient
_httpClient
_syncRoot

CSharp/Library/Microsoft.Bot.Connector.Shared/ConnectorClientEx.cs
```diff
@@ -49,25 +49,44 @@ public ConnectorClient(Uri baseUri, MicrosoftAppCredentials credentials, HttpCli
...
+                        g_httpClient.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue("Microsoft-BotFramework", "3.1"));
+                        g_httpClient.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue($"(BotBuilder .Net/{GetClientVersion(this)})"));
+                        g_httpClient.DefaultRequestHeaders.ExpectContinue = false;
+                    }
+                }
+            }
+
+            // use global singleton for perf reasons
+            this.instanceClient = this.HttpClient;
+            this.HttpClient = g_httpClient;
```
cleemullins: This is a big hack:
1. Creates a static global, in a lock. Clear 'nuf.f
2. Preserve the previous httpClient. 
3. Swap in the global one. 
4. Later, in the Dispose, swap the instance one back in to be disposed. 

Do this weird behavior w/o comments in the code as to whyat's going on and why. 

There's got to be something cleaner.

CSharp/Library/Microsoft.Bot.Connector.Shared/StateClientEx.cs
```diff
@@ -97,9 +100,44 @@ private static DelegatingHandler[] AddJwtTokenRefresher(DelegatingHandler[] srcH
...
+                    {
+                        g_httpClient = new HttpClient();
+                        g_httpClient.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue("Microsoft-BotFramework", "3.1"));
+                        g_httpClient.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue($"(BotBuilder .Net/{GetClientVersion(this)})"));
+                        g_httpClient.DefaultRequestHeaders.ExpectContinue = false;
+                    }
+                }
+            }
+
+            // use global singleton for perf reasons
```
cleemullins: It's that hack again, of swapping one in and disposing it. 

Note: You could inherit from HttpClient and create a NonDisposableHttpClient, that just ignores Dispose.

PR #4413, opened 2018-04-02
Remove GetStateClient method

JasonSowers:
This method has been the source of much confusion for customers.  Since the migration deadline has "passed"  it is probably best to remove this in versions moving forward


xjose97x:
Was looking forward for this.
But what about a GetStateClient that gets the state client provided by the user?
I understand the issues we had with this was because this method retrieved the bot framework default (now deprecated) state client, right? 
What if we modify it to retrieve the user provided state client?

I believe that is very useful in scenarios where you want to access the user/conversation data from a place where you don't have access to the IDialogContext and only the activity (for instance, the MessageController in c#)

JasonSowers:
I was thinking about this too, I'll talk it over with the team.  Thanks for the input @xjose97x

EricDahlvang:

PR #4424, opened 2018-04-05
Fix for unseekable stream + typo fix

HristinaPopova:
Attachments API test cases were failing because AttachmentsExtension was throwing unseekable stream.

Also, fixing a typo in a /// comment block (which is included in 2 other auto-generated files).


CSharp/Library/Microsoft.Bot.Connector.Shared/ConnectorAPI/AttachmentsExtensions.cs
```diff
@@ -95,9 +95,8 @@ public static byte[] GetAttachment(this IAttachments operations, string attachme
             using (var _result = await operations.GetAttachmentWithHttpMessagesAsync(attachmentId, viewId, null, cancellationToken).ConfigureAwait(false))
             {
                 MemoryStream memoryStream = new MemoryStream();
```
willportnoy: >MemoryStream [](start = 48, length = 12)

fyi, MemoryStream is Disposable

PR #4425, opened 2018-04-05
Tomlm/update swagger2

tomlm:
* New autorest generated enums
* Trace Activity
* new GDPR API's
* AuthSSO definitions


ToddN:
>                 "consumes": [

why is the empty space important?

---
Refers to: CSharp/Library/Microsoft.Bot.Connector.Shared/Swagger/ConnectorAPI.json:31 in 9e6d339. [](commit_id = 9e6d339e17aeed73da7d402fb446ad9eab8bd0fe, deletion_comment = False)

CSharp/Library/Microsoft.Bot.Connector.Shared/ConnectorAPI/Models/ActionTypes.cs
```diff
@@ -0,0 +1,27 @@
+// <auto-generated>
```
ToddN: Won't the lines in generateClient.cmd (in BotBuilder) that say "call ..\node_modules\.bin\replace "using Models;" "" . -r --include="*.cs""  cause build errors if we move these to Models?   Should we update that script as well? Or leave these back in the ConnectorAPI directory?

PR #4509, opened 2018-04-22
updated the LUIS TryFindEntity extension to only return an entity by …

gtewksbury:
…type if its start index doesn't overlap with the start and end index of other LuisResult Entities.  I came across this issue when working with an intent that accepts both a date and number.  If the intent message is 'make a reservation tomorrow morning at 2 for 7 people', 7 being the trained 'number' in LUIS, the current TryFindEntity extension was pulling back '2', which was already considered part of the datetime entity.


JasonSowers:
Thank you for your contribution.  Could you please provide the tests you used while making this addition?

gtewksbury:
Hey Jason, I added unit tests and made a small correction in the latest commit.

Issue #3926, opened 2017-12-14
[Node.js SDK, bug] v3.13.1 release throws error on incoming messages


harshithkashyap:
## Bot Info 
* App ID: 
* SDK Platform: Node.js
* SDK Version: 3.13.1
* Active Channels: DirectLine
* Deployment Environment: Azure app service

## Issue Description
After updating the SDK to 3.13.1, the botbuilder SDK is throwing the following error.
```
Error: Can't set headers after they are sent.
```

It seems like [this commit](https://github.com/Microsoft/BotBuilder/commit/eca181033aa688b3906ef249cf1315420a291176) introduced the change adding a `defaultNext`.

We are using botbuilder with express.js. v3.12 works without any issues.

matthewfcarlson:
I'm also seeing this and have a consistent repo.

harshithkashyap:
@nwhitmont  This issue still exists with botbuilder@3.14. Any chance this could get fixed in the next release as we are missing out on SDK updates?

daveta:
Thank you for opening an issue against the Bot Framework SDK v3. As part of the Bot Framework v4 release, we’ve moved all v3 work to a new repo located at https://github.com/microsoft/botbuilder-v3. We will continue to support and offer maintenance updates to v3 via this new repo.
 
From now on, https://github.com/microsoft/botbuilder repo will be used as hub, with pointers to all the different SDK languages, tools and samples repos.
 
As part of this restructuring, we are closing all tickets in this repo.
 
For defects or feature requests, please create a new issue in the new Bot Framework v3 repo found here:
https://github.com/microsoft/botbuilder-v3/issues


For Azure Bot Service Channel specific defects or feature requests (e.g. Facebook, Twilio, Teams, Slack, etc.), please create a new issue in the new Bot Framework Channel repo found here:
https://github.com/microsoft/botbuilder-channels/issues
 
For product behavior, how-to, or general understanding questions, please use Stackoverflow.
https://stackoverflow.com/search?q=bot+framework


Thank you.
 
The Bot Framework Team

PR #4610, opened 2018-05-15
fix(ChatConnector): remove next() calls from dispatch on success operations

harshithkashyap:
Partially reverts #3193 

Express/Restify middlewares after the `/api/messages` definition should only be called in case of errors. As of now, the `next()` function is called on all HTTP 200 outgoing requests which leads to ExpressJS calling 404 error handler even on success operations.

```javascript
app.post('/api/messages', jwtCheck, connector.listen());

// catch 404 and forward to error handler
app.use((req, res, next) => {
    const err = new Error('Not Found');
    err.status = 404;
    next(err);
});
```

The above 404 error handler or any other subsequent middleware should only be called if none of the endpoints match the incoming request.

Relates to #3926 

Published under `botbuilder-tmp` to test.
```bash
npm i -E botbuilder-tmp
```

### DIFF
--- a/Node/core/lib/bots/ChatConnector.js
+++ b/Node/core/lib/bots/ChatConnector.js
@@ -561,12 +561,10 @@ var ChatConnector = (function () {
                 else if (body) {
                     res.send(status || 200, body);
                     res.end();
-                    next();
                 }
                 else {
                     res.status(status || 200);
                     res.end();
-                    next();
                 }
             });
         }
--- a/Node/core/package-lock.json
+++ b/Node/core/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "botbuilder",
-  "version": "3.14.1",
+  "version": "3.15.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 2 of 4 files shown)


bilby91:
Hey @harshithkashyap , can you point to the documentation in `express` that describes your comment "The above 404 error handler or any other subsequent middleware should only be called if none of the endpoints match the incoming request." 

Thanks!

harshithkashyap:
@bilby91 https://expressjs.com/en/guide/using-middleware.html
> If the current middleware function does not end the request-response cycle, it must call next() to pass control to the next middleware function. Otherwise, the request will be left hanging.

Since we are ending the response, I skipped the `next()` calls on success operations.
Also note that 404 error handlers are defined as regular middleware with 3 arguments (https://expressjs.com/en/starter/faq.html). So, `next()` calls 404 middleware and ends up setting the response as HTTP 404 even on 200.

harshithkashyap:
@Stevenic Could you have a look at this when you get a chance and merge the PR? This PR has been open for more than 2 months now.

EricDahlvang:
Thank you for re-submitting this on the new repository.

PR #4637, opened 2018-05-21
Add support for OAuthCards when there is no AppId or Pwd

Jeffders:
This change includes:

- Support for 'emulating OAuthCards' when using the emulator (so no remote hosts are involved)
- Support for using OAuthCards when there is no AppId + Pwd in the emulator (in which case the SDK uses the emulated OAuthCards)

### DIFF
--- a/CSharp/Library/Microsoft.Bot.Connector.Shared/OAuthAPI/OAuthApi.cs
+++ b/CSharp/Library/Microsoft.Bot.Connector.Shared/OAuthAPI/OAuthApi.cs
@@ -96,9 +96,7 @@ namespace Microsoft.Bot.Connector
             HttpResponseMessage _httpResponse = null;
             _httpRequest.Method = new HttpMethod("GET");
             _httpRequest.RequestUri = new System.Uri(_url);
-
-            MicrosoftAppCredentials.TrustServiceUrl(_url);
-
+            
             // Set Credentials
             if (Client.Credentials != null)
             {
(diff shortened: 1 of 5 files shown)


CSharp/Library/Microsoft.Bot.Connector.Shared/OAuthAPI/OAuthApi.cs
```diff
@@ -297,7 +291,7 @@ public Task<TokenResponse> GetUserTokenAsync(string userId, string connectionNam
             return String.Empty;
         }
 
-        public async Task SendEmulateOAuthCardsAsync(bool emulateOAuthCards)
+        public async Task SendEmulateOAuthCardsAsync(string emulatorUri, bool emulateOAuthCards)
```
tomlm: >SendEmulateOAuthCardsAsync(string emulatorUri, bool emulateOAuthCards) [](start = 26, length = 70)

isn't this a breaking API change?

cleemullins:
@Jeffders Hey Jeff, is this something ready to be merged?

CSharp/Library/Microsoft.Bot.Connector.Shared/OAuthAPI/OAuthApi.cs
Jeffders: Yes, good point. Here's how I fixed this: I reverted this part in favor of making a new oauth client that points just at the emulator. See latest changes.

PR #4693, opened 2018-06-04
[Node.js] return `this` instead of `Message`

m-shaka:
The method chain causes compile error because `Message.text` returns `Message` and `addMentionToText` is a member of `TeamsMessage` that is subtype of `Message`.

```typescript
import builder = require('botbuilder')
import teams = require('botbuilder-teams')

const bot = new builder.UniversalBot(connector, session => {
    const user: builder.IIdentity = {id: '1', name: 'John Doe'}
    const msg = new teams.TeamsMessage(session)
      .text('text')
      .addMentionToText(user)
    session.send(msg)
})
```


EricDahlvang:

PR #4923, opened 2018-08-01
Set all fields to null on 'reset' command in formflow

nyxsys:
Tested this with the formflowbot locally without issue, this code only runs when the "reset" command is given by a user and only resets steps marked as 'field' within a form, this results in the desired functionality for "reset" where all fields are reprompted to the user from the start.

Previously when a 'reset' command was issued the bot would reset the formState but leave the internal state alone, meaning that the values entered into it would still persist. This was meant for them to act as default values for the form, but when the first field was reprompted and an input was entered the other states would evaluate to "completed" due to the "skipstep" function. This fix avoids that by just setting all fields back to null (their original value) when reset is issued.


Rufus1123:
Any updates on when or whether this will be merged?

EricDahlvang:

PR #4937, opened 2018-08-04
Make the IgnoreField and Order attributes work on properties as well as fields

### LINKED ISSUE #4921: IgnoreFieldAttribute and OrderAttribute should work on properties
When constructing a FormFlow model, the `IgnoreField` attribute can be placed on both fields and properties, but seems to have no effect on the latter. So either honor the attribute or disable placement on properties and make a new `IgnoreProperty` attribute for that purpose.

FilipFilipov:
Fixes #4921


EricDahlvang:

PR #4959, opened 2018-08-10
Languageresources finnish

OskariRomanoff:
Added .xlf and .resx language resource files for finnish language


cleemullins:
This isn't a repo we've been localizing. Once we do, we'll do everything at once.

Issue #2409, opened 2017-03-09
Date input is printed incorrectly


maba4891:
Hi,

in my node.js bot I provide a simple builder.Prompts.text to ask the user for a date (I know that Prompts.time exists but since it does not seem to work with German dates yet I use Prompts.text for now and parse the date with the datejs module).

The problem is that if the user enters the date in a German format (e.g. "12. Februar 2017") and presses enter he sees "1. Februar 2017" as his input. The same is true for other dates (e.g. "20. August 2016") which are printed in a similar way like "1. August 2016". This also happens when I provide a date with a dot as an input to any other type of prompt.

Somehow the **dot** seems to be the problem here, 12 Februar 2017 works just fine (as does 12.2.2017). 

The funny thing is also that **internally the right date is saved**, e.g. 12. Februar 2017 even if 1. Februar 2017 is printed (I save the input in session.userData). So it seems the problem is just the way the input is printed.

Any clue what is happening here?

maba4891:
Edit: This problem seems to be only affecting the Emulator and the Web Chat, on Skype and Facebook Messenger everything seems to be fine.

danmarshall:
Thanks for filing this issue @maba4891 . Would you mind filing this issue on the WebChat repo?
https://github.com/Microsoft/BotFramework-WebChat/issues

maba4891:
@danmarshall no problem, I'll post it there

danmarshall:
Thanks!

PR #4986, opened 2018-08-18
Add forwardDate and locale options to IPromptRecognizeTimesOptions

GhassenRjab:
builder.Prompts.time doesn't recognize German dates like 07.11 (7th of November) or Morgen (tomorrow)
I have added a locale option to IPromptRecognizeTimesOptions in order to use a localized version of chrono in order to improve accuracy of parsing the dates.

Issue #2409 had the same problem but opted to use builder.Prompts.text and parse the utterance using another library. A solution i couldn't follow.

Also, I added a forwardDate option too because i think it might come handy in some cases.

Two issues I have encountered when making the changes are:
First, the transpiler didn't work unless I changed the import of chrono-node from this
`import * as chrono from 'chrono-node';`
to this
`const chrono = require('chrono-node');`
because I am using this statement
`results = chrono[locale].parse(utterance, refDate, { forwardDate });`
Second, after transpiling from TypeScript to JavaScript, all JS files had changes I didn't make in
`var extendStatics`

### DIFF
--- a/Node/core/lib/RemoteSessionLogger.js
+++ b/Node/core/lib/RemoteSessionLogger.js
@@ -1,8 +1,11 @@
 "use strict";
 var __extends = (this && this.__extends) || (function () {
-    var extendStatics = Object.setPrototypeOf ||
-        ({ __proto__: [] } instanceof Array && function (d, b) { d.__proto__ = b; }) ||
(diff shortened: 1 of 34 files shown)


EricDahlvang:
Closing this PR because the code for this repository has been moved to https://github.com/Microsoft/BotBuilder-V3  If you think this PR is still relevant, please re-submit on the new repository.  (If you re-submit, please do not include the no-op changes.)

PR #4988, opened 2018-08-19
[Botbuilder SDK for node.js] Constructor of WaterfallDialog updated

jKelio:
I changed the Array.isArray(steps) condition to steps instanceof Array to ensure that the array could also be in the prototyping chain like inheritance.

I hope this change would be okay for you. It will now allow us to use an object with an Array Prototype in its prototyping chain?

Thank you and best regards from Augsburg,
Leon


jKelio:
This change is for the node.js SDK

PR #5037, opened 2018-09-05
fix objectDisposed exception in RefreshTokenAsync()

tomlm:
HttpResponseMessage implements IDispose, which internally calls .Dispose() on the HttpContent passed into PostAsync(), invalidating the content that was attempted to be reusued in another request.

This fix creates a new Content to be submitted on every attempt by moving the FormUrlEncodedContent inside the retry loop

### DIFF
--- a/CSharp/Library/Microsoft.Bot.Connector.Shared/MicrosoftAppCredentials.cs
+++ b/CSharp/Library/Microsoft.Bot.Connector.Shared/MicrosoftAppCredentials.cs
@@ -310,7 +310,9 @@ namespace Microsoft.Bot.Connector
         /// <returns></returns>
         private async Task<OAuthResponse> RefreshTokenAsync()
         {
-            var content = new FormUrlEncodedContent(new Dictionary<string, string>()
+            while (true)
+            {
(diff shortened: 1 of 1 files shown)


CSharp/Library/Microsoft.Bot.Connector.Shared/MicrosoftAppCredentials.cs
```diff
@@ -336,13 +336,13 @@ private async Task<OAuthResponse> RefreshTokenAsync()
...
 #else
-                        logger?.LogWarning($"RefreshTokenAsync() Exception) {body ?? response.ReasonPhrase}");
+                        logger?.LogWarning($"RefreshTokenAsync() Exception {body ?? response.ReasonPhrase}");
 #endif
 
                     }
 
                     // try again in a bit to prevent hammering a service if it's not working
-                    await Task.Delay((int)TimeSpan.FromSeconds(1).TotalMilliseconds);
+                    await Task.Delay(TimeSpan.FromSeconds(30));
```
EricDahlvang: 30 seconds?  Is this change intentional? (that's a long time)

PR #5058, opened 2018-09-17
Corrected a small bug 

NicovdijkWork:
I know I'm supposed to first log an issue etc. but it's a really small thing. On line 172 in , brackets were missing in the logging for WaterfallDialog.


sgellock:
order of precedence is correct.  the brackets aid in readability.  You're call @Stevenic

NicovdijkWork:
I do not think the order of precedence is correct. Try this:
`"use strict";

var step = 1;
var steps = {length: 2};
console.log('With brackets:\nwaterfall() step ' + (step + 1) + ' of ' + steps.length);
console.log('Without brackets:\nwaterfall() step ' + step + 1 + ' of ' + steps.length);`

My output is:
`With brackets:
waterfall() step 3 of 2
Without brackets:
waterfall() step 21 of 2`

PR #5073, opened 2018-09-23
Revised entity design

dandriscoll:
Update Activity schema, manifest spec, and examples to reflect revised way of passing entities into actions.


dandriscoll:
What a mess

PR #5089, opened 2018-10-02
Fix link for license file

cseas:
Previous link directed to license file for vscode. Fixed it.


stevengum:
Thank you @cseas for catching this!

PR #5289, opened 2019-02-07
tokenParameters are not required when creating a conversation

EricDahlvang:


dandriscoll:
This file is generated but the update matches what we have on our side.

PR #5664, opened 2019-11-04
Add json file for alternate UI & cleanup

daveta:
- @compulim going to consume a JSON file to improve the user experience
- Cleanup to isolate output and helpers.


dri/report.py
```diff
@@ -56,11 +63,13 @@
     'azure/azure-cli',
 ]
 
+# Do not apply user filters to these repos.
 BYPASS_USERFILTER_REPOS = [
     'botbuilder-tools',
 ]
 
-MICROSFT_EMPLOYEES=[
```
tsuwandy: Perhaps we rename this to MSFT_EMPLOYEES so it doesn't look like a misspell? :-)

PR #5682, opened 2019-12-03
Updated with BF Team handles

gabog:
Updated python file with BF team users


gabog:
Need to check with @daveta if this is really needed, closing the PR for now.

Issue #5592, opened 2019-09-09
Define skills 2.1 manifest schema


gabog:
A skill should expose its contract and supported parameters so it can be consumed and added to a parent bot. 

It should handle versioning (for the manifest format and for the skill itself)

It should describe return values from the skill.

A skill can optionally expose an endpoint that returns a manifest defining the contract to invoke the skill. 

Here is an example of the manifest currently exposed by the VA calendar skills as a reference:

```
{
  "id": "calendarSkill",
  "msaAppId": "f67037d3-3cb9-4a96-820c-bd3ee5e8c06e",
  "name": "Calendar Skill",
  "endpoint": "http://bfcalendarskill.azurewebsites.net/api/skill/messages",
  "description": "The Calendar skill provides calendaring related capabilities and supports Office and Google calendars.",
  "iconUrl": "http://bfcalendarskill.azurewebsites.net/calendarSkill.png",
  "authenticationConnections": [
    {
      "id": "Outlook",
      "serviceProviderId": "Azure Active Directory v2",
      "scopes": "User.ReadBasic.All, Calendars.ReadWrite, People.Read, Contacts.Read"
    },
    {
      "id": "Google",
      "serviceProviderId": "Google",
      "scopes": "https://www.googleapis.com/auth/calendar https://www.googleapis.com/auth/contacts"
    }
  ],
  "actions": [
    {
      "id": "calendarskill_createEvent",
      "definition": {
        "description": "Create a new event",
        "slots": [
          {
            "name": "title",
            "types": [
              "string"
            ]
          },
          {
            "name": "content",
            "types": [
              "string"
            ]
          },
          {
            "name": "attendees",
            "types": [
              "string"
            ]
          },
          {
            "name": "startdate",
            "types": [
              "string"
            ]
          },
          {
            "name": "starttime",
            "types": [
              "string"
            ]
          },
          {
            "name": "duration",
            "types": [
              "string"
            ]
          },
          {
            "name": "location",
            "types": [
              "string"
            ]
          },
          {
            "name": "timezone",
            "types": [
              "string"
            ]
          }
        ],
        "triggers": {
          "utteranceSources": [
            {
              "locale": "en",
              "source": [
                "Calendar#CreateCalendarEntry",
                "Calendar#FindMeetingRoom"
              ]
            }
          ]
        }
      }
    },
    {
      "id": "calendarskill_changeEventStatus",
      "definition": {
        "description": "Change the status of an event (accept/decline).",
[...]

PR #5693, opened 2019-12-11
Skill manifest schema 2.1 preview 1 and sample updates

gabog:
Relates to #5592 

This PR is a preview for the skill manifest schema 2.1. It includes the follow changes:

- Adds a dispatchModels property for defining lu modes and intent mappings to activities.
- Adds a activitiesSent property for defining proactive messages that can be sent by a bot

To review, the easiest thing is to probably look at [the sample](https://github.com/microsoft/botframework-sdk/blob/gabog/SkillSchema21/schemas/skills/samples/complex-skillmanifest-2.1.preview-0.json) and see if that meets your needs. 

You can also check the schema and descriptions for each new property.

Please review and provide any comments.


schemas/skills/samples/complex-skillmanifest-2.1.preview-0.json
```diff
@@ -0,0 +1,184 @@
...
+        }
+    ],
+    "dispatchModels": {
+        "languages": {
+            "en": [
+                {
+                    "id": "SkillLuModel-en",
+                    "name": "SkillBot LU (English)",
+                    "contentType": "application/lu",
+                    "url": "http://sample.com/SkillBot-en.lu",
```
lzc850612: so in reality, where should a developer put the lu file? is it hosted together with the skill bot?

schemas/skills/samples/complex-skillmanifest-2.1.preview-0.json
```diff
@@ -0,0 +1,184 @@
...
+                },
+                "postalCode": {
+                    "type": "string",
+                    "title": "Postal code"
+                }
+            }
+        }
+    },
+    "activitiesSent": {
+        "flightUpdated": {
```
lzc850612: is this the proactive message support where flightUpdated is the event's name that a parent bot should be looking for?

schemas/skills/samples/complex-skillmanifest-2.1.preview-0.json
gabog: The developer can chose to put the lu file wherever the feel like it, it could be on the same place where the bot is hosted or somewhere else. In other words, we are not prescribing where these files are. Makes sense?

schemas/skills/samples/complex-skillmanifest-2.1.preview-0.json
gabog: Right, this section defines the proactive messages a parent should expect from a child bot.

schemas/skills/samples/complex-skillmanifest-2.1.preview-0.json
```diff
@@ -1,5 +1,5 @@
 {
-    "$schema": "https://raw.githubusercontent.com/microsoft/botframework-sdk/master/schemas/skills/skill-manifest-2.1.preview-0.json",
+    "$schema": "https://schemas.botframework.com/schemas/skills/skill-manifest-2.1.preview-0.json",
     "$id": "SkillBot",
     "name": "Sample skill definition that can handle multiple types of activities",
```
darrenj: One thing the Power-VA team raised was the need to localise fields like name/description - there's a balance in my view between a manifest describing the service for purposes of communicating and something like a store/directory registration

schemas/skills/samples/complex-skillmanifest-2.1.preview-0.json
darrenj: We do have a LUIS URI format (same for speech priming) that would enable interrogation through the authoring API (with an authoring key - now retrievable automatically) to extract utterances. I assume by this the URL is being kept open/flexible so could use what ever you like?

schemas/skills/samples/complex-skillmanifest-2.1.preview-0.json
```diff
@@ -0,0 +1,184 @@
...
+            "name": "eu",
+            "protocol": "BotFrameworkV3",
+            "description": "Production  endpoint for SkillBot in Europe",
+            "endpointUrl": "http://myskill.contoso.com/api/messages",
+            "msAppId": "11111111-0000-0000-0000-000000000000"
+        }
+    ],
+    "dispatchModels": {
+        "languages": {
+            "en": [
```
darrenj: picky ;-) `en-us` ?

schemas/skills/samples/complex-skillmanifest-2.1.preview-0.json
```diff
@@ -0,0 +1,184 @@
...
+                {
+                    "id": "SkillLuModel-en",
+                    "name": "SkillBot LU (English)",
+                    "contentType": "application/lu",
```
darrenj: where are these types defined?

schemas/skills/samples/complex-skillmanifest-2.1.preview-0.json
```diff
@@ -0,0 +1,184 @@
...
+                    "url": "http://sample.com/SkillBot-es-MX.lu",
+                    "description": "Spanish (Mexico) language model for the skill"
+                }
+            ]
+        },
+        "intents": {
+            "bookFlight": "#/activities/bookFlight",
+            "getWeather": "#/activities/getWeather",
+            "*": "#/activities/message"
```
darrenj: Intents relate to what we had as actions before? So there are 2 - bookflight and getweather? * means anything else will just go to the root dialog for processing directly - as if the user typed a message directly to the bot?

schemas/skills/samples/complex-skillmanifest-2.1.preview-0.json
```diff
@@ -0,0 +1,184 @@
...
+        }
+    },
+    "activities": {
+        "bookFlight": {
+            "description": "Books a flight",
+            "type": "event",
```
darrenj: What is the different between an event or an invoke (below?) beyond the activitytype ;-). Are we recommending one over the other?

schemas/skills/samples/complex-skillmanifest-2.1.preview-0.json
```diff
@@ -0,0 +1,184 @@
...
+                "$ref": "#/definitions/weatherReport"
+            }
+        },
+        "message": {
+            "type": "message",
+            "description": "Receives the user's' utterance and attempts to resolve it using the skill's LU models"
+        }
+    },
+    "definitions": {
+        "localeValue": {
```
darrenj: localeValue not referenced anywhere?

schemas/skills/samples/complex-skillmanifest-2.1.preview-0.json
```diff
@@ -0,0 +1,184 @@
...
+            "properties": {
+                "locale": {
+                    "type": "string",
+                    "description": "The current user's locale ISO code"
+                }
+            }
+        },
+        "bookingInfo": {
+            "type": "object",
+            "required": [
```
darrenj: would we not put required on the individual properties rather than seperately?

schemas/skills/samples/complex-skillmanifest-2.1.preview-0.json
```diff
@@ -0,0 +1,184 @@
...
+                "origin"
+            ],
+            "properties": {
+                "origin": {
```
darrenj: what types/primitivies are supported?

schemas/skills/samples/complex-skillmanifest-2.1.preview-0.json
```diff
@@ -0,0 +1,184 @@
...
+                "postalCode": {
+                    "type": "string",
+                    "title": "Postal code"
+                }
+            }
+        }
+    },
+    "activitiesSent": {
+        "flightUpdated": {
+            "type": "event",
```
darrenj: as above - are events the proposed way forward or invoke?

schemas/skills/samples/complex-skillmanifest-2.1.preview-0.json
darrenj: We also use the suffix of #{IntentName} on a LU file to enable you to reference specific intents and not expose everything within an LU file. For example this is how we were mapping intent(s) to a given action, in the newly proposed model this is mapped to the skill level not individual actions (which is fine) but wonder how we might enable a user to reference a LU file but only certain intents.  We could impose the LU file provided is cut down to just what's needed for skill purposes from the main LU file but wanted to confirm

```
    "utteranceSources": [
            {
              "locale": "en-us",
              "source": [
                "Calendar#CreateCalendarEntry",
                "Calendar#FindMeetingRoom"
              ]
            }
```

schemas/skills/samples/complex-skillmanifest-2.1.preview-0.json
gabog: An event uses http back and forth (like we normally use). 

An invoke is request/reply, sort of a function call, it is always single turn and the caller will get some return value in the response object.

There is a sample invoke in the [DialogToDialog](https://github.com/microsoft/BotBuilder-Samples/blob/master/experimental/skills/DialogToDialog/DialogRootBot/Dialogs/MainDialog.cs#L139)  sample

schemas/skills/samples/complex-skillmanifest-2.1.preview-0.json
gabog: The ones defined in [json-schema.org draft 07](https://json-schema.org/understanding-json-schema/reference/index.html)

schemas/skills/samples/complex-skillmanifest-2.1.preview-0.json
gabog: The most common one will be event (and the one I will personally mostly recommend), Teams uses invokes in some situations but they are very uncommon.

schemas/skills/samples/complex-skillmanifest-2.1.preview-0.json
gabog: We don't have constants for it now, we are leaving it open for now, sort of like with http requests content types where you have well known ones but you can put whatever you want.
Once we build a manifest parser we may be more strict but for now, we are assuming everything is lu files so the only one I know off is application/lu.

schemas/skills/samples/complex-skillmanifest-2.1.preview-0.json
darrenj: Great - so invokes would not return any activities / messages during processing? If so makes sense. Do you know if PVA plan to support invokes?

PR #5705, opened 2020-01-06
4.7 update notes

christopheranderson:
- update 4.7 version where appropriate
- adjust language on the preview features section


christopheranderson:
We might want to just call that section "What's new" instead of Preview. I'm gonna avoid making further changes until we straighten out the botframework vs botframework-sdk stuff.

PR #5806, opened 2020-04-16
move component schema to botframework-sdk

tomlm:
* Move component.schema 
* Move baseComponent.schema 
* add readme.txt


schemas/component/readme.md
```diff
@@ -0,0 +1,6 @@
+# Component Schema
+The component schema define extensions to JsonSchema for supporting
+Bot Framework declarative component definitions (aka .dialog files).
```
chrimc62, line 3: Should be .schema.

PR #5831, opened 2020-05-07
Skills manifest: Add readme & more complexity to the sample

clearab:
In this PR:

- Add some "other" activities to the complex skill manifest for testing purposes
- add a README to the skills folder

### DIFF
--- a/schemas/skills/readme.md
+++ b/schemas/skills/readme.md
@@ -1,0 +1,11 @@
+# Skills JSON schema
+
+This folder contains the current and previous versions of the JSON Schema that defines your skill manifest. For additional information on skills, skill consumers, and skill manifests see the [skills documentation](https://docs.microsoft.com/azure/bot-service/skills-conceptual?view=azure-bot-service-4.0).
+
+## Published location
+
+The skills JSON schema is published at:
+
+`https://schemas.botframework.com/schemas/skills/{file name}`
+
+You should use the published version when referencing this schema.
--- a/schemas/skills/samples/complex-skillmanifest-2.1.preview-1.json
+++ b/schemas/skills/samples/complex-skillmanifest-2.1.preview-1.json
@@ -111,13 +111,7 @@
             "type": "typing"
         },
         "conversationUpdate": {
-            "type": "conversationUpdate",
-            "name": "memberAdded",
-            "value": {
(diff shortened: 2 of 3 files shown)


schemas/skills/samples/complex-skillmanifest-2.1.preview-1.json
```diff
@@ -106,6 +106,18 @@
         "message": {
             "type": "message",
             "description": "Receives the user's' utterance and attempts to resolve it using the skill's LU models"
+        },
+        "typing": {
+            "type": "typing"
+        },
+        "conversationUpdate": {
+            "type": "conversationUpdate",
+            "name": "memberAdded",
```
gabog, line 115: I think we need to remove this, conversationUpdate doesn't have a name property set (it is null by default)

schemas/skills/samples/complex-skillmanifest-2.1.preview-1.json
```diff
@@ -106,6 +106,18 @@
         "message": {
             "type": "message",
             "description": "Receives the user's' utterance and attempts to resolve it using the skill's LU models"
+        },
+        "typing": {
+            "type": "typing"
+        },
+        "conversationUpdate": {
```
gabog, line 113: Not sure if this truly represents a conversationUpdate activity, to be able to cast this to an OOTB SDK type it should represent https://github.com/microsoft/botbuilder-dotnet/blob/master/libraries/Microsoft.Bot.Schema/IConversationUpdateActivity.cs.
Value is not set for conversationUpdates for example.

schemas/skills/readme.md
```diff
@@ -0,0 +1,11 @@
+# Skills JSON schema
+
+This folder contains the current and previous versions of the JSON Schema that defines your skill manifest. For additional information on skills, skill consumers, and skill manifests see the [skills documentation](https://docs.microsoft.com/azure/bot-service/skills-conceptual?view=azure-bot-service-4.0).
```
gabog, line 3: Consider creating an aka.ms alias here in case we move the docs so we don't need to do a new PR to update this. 
@Kaiqb, knows the alias of the team that needs to be added to the aka.ms entry to tech writers and see it and update it if needed.

PR #5833, opened 2020-05-08
Finalize skill manifest versioning

clearab:
This PR finalizes the skill manifest versioning process that should accommodate and simplify future versions of the manifest, while remaining backwards compatible.


clearab:
Easier to view by looking at the branch directly: https://github.com/microsoft/botframework-sdk/tree/clearab/skills-manifest-version/schemas/skills

PR #5954, opened 2020-07-07
Update component schema to add $role expression to `expression` and `equalsExpression` definitions

vishwacsena:
Fixes microsoft/botbuilder-dotnet#3839

Composer uses `$role:expression` to determine if a particular property should be rendered with an expression editor UI control. At the moment certain properties that carry `$ref:schema#/definitions/expression` as well as `$ref:schema:#/definitions/equalsExpression` are not rendered as expression controls in composer UI. 

This PR updates the component schema to add the missing `$role: expression` to the definitions.

### DIFF
--- a/schemas/component/definitions.schema
+++ b/schemas/component/definitions.schema
@@ -33,6 +33,7 @@
             }
         },
         "expression": {
+            "$role": "expression",
             "type": "string",
             "title": "Expression",
             "pattern": "^.*\\S.*",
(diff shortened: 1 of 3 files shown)


schemas/component/definitions.schema
```diff
@@ -33,6 +33,7 @@
             }
         },
         "expression": {
+            "$role": "expression",
             "type": "string",
```
chrimc62, line 37: This won't help.  The file is not actually used and all of the definitions are in component.schema which is where this change should go.  Gabo has a PR to move this.

vishwacsena: Updated the new V1 definitions checked in by @gabog

PR #5955, opened 2020-07-07
add ui.schema

a-b-r-o-w-n:
Adds ui.schema json schema to define customization options for SDK components.


schemas/component/ui.schema
```diff
@@ -0,0 +1,113 @@
...
+  "patternProperties": {
+      "^\\$schema$": {
+          "type": "string"
+      },
+      "^(?!\\$role:).+$": {
+          "title": "$kind",
+          "description": "A map of component names to UI options with customizations for each component.",
+          "$ref": "#/definitions/UIOptionsRecursive"
+      },
+      "^\\$role:.+$": {
```
a-b-r-o-w-n, line 16: This may not be necessary to expose at the SDK level and may be more of a Composer only customization. We use this mechanism to provide some custom UI for expressions.

chrimc62: Seems like I can achieve the same effect on an individual property.  This is a "global" default.  Today we do not allow $roles that we did not define so this really boils down to the question of allowing someone to customize all properties with a $role.  I'm not sure if that is worth it or not.  

---
In reply to: [451157108](https://github.com/microsoft/botframework-sdk/pull/5955#discussion_r451157108) [](ancestors = 451157108)

cwhitten:
Let's add how to model, customize, and extend the menu flyouts and continue to riff on the scenarios to customize node views for custom actions.

schemas/component/ui.schema
```diff
@@ -0,0 +1,113 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "https://raw.githubusercontent.com/microsoft/botframework-sdk/master/schemas/component/ui.schema",
```
chrimc62, line 3: >https://raw.githubusercontent.com/microsoft/botframework-sdk/master/schemas/component/ui.schema [](start = 10, length = 95)

Gabo is about to check-in that this would change to: [https://schemas.botframework.com/schemas/component/v1.0/ui.schema](https://schemas.botframework.com/schemas/component/v1.0/ui.schema)
You check it into schemas/component/v1.0/ and it gets mapped to that URL.

cwhitten:
Notes from chat:

SendActivity.ui.schema
```json
{
  "menu": {}
}
```

CustomComponent.ui.schema
```json
{
  "menu": {
     "foo": true
   }
}
```

yields the following merged menu:

[SendActivity (leaf)]
[Foo -> Custom Component (leaf)]

schemas/component/ui.schema
a-b-r-o-w-n: I think I am going to remove this and only expose it through the JS interface for the moment, if at all.

a-b-r-o-w-n:
@cwhitten what about something like this?

SendActivity.ui.schema
```json
{
  "menu": {
    "label": "Send a response"
  }
}
```

MyAction.ui.schema
```json
{
  "menu": {
    "label": "My Awesome Action",
    "group": ["My Company"]
  }
}
```

OtherAction.ui.schema
```json
{
  "menu": {
    "group": ["My Company", "Other Category"]
  }
}
```

Produces
```
[Send a response]
[My Company]
  -> [My Awesome Action]
  -> [Other Category]
     -> [OtherAction]
```

schemas/component/ui.schema
a-b-r-o-w-n: Ok. I'll update once that has been merged.

cwhitten:
Looks good.

a-b-r-o-w-n:
@zxyanliu Adding you as a reviewer to do a pass on naming / copy.

schemas/component/ui.schema
```diff
@@ -0,0 +1,113 @@
...
+                          "type": "string"
+                      },
+                      {
+                          "const": false
+                      }
+                  ]
+              },
+              "order": {
+                  "title": "Property Order",
+                  "description": "Set order of fields. Use \"*\" for all other fields. ex. [\"foo\", \"*\", \"bar\"]",
```
zxyanliu, line 59:
```suggestion
                  "description": "Set the order of fields. Use \"*\" for all other fields. ex. [\"foo\", \"*\", \"bar\"]",
```

zxyanliu: The naming and descriptions look good to me.

cwhitten:
A principle I’m considering for the flow section: What goes in the ui component schema are obvious properties of the view that closely represent the hierarchy of general information and behavior we want to expose from the dialog’s component schema and implementation.

Additionally, we already have a concept of sections in a node that models with an imperative language interface if you squint right: 
head, body, footer : parameters, implementation, return value. A lot of the copy in these sections map to the information we are trying to convey.

With this principle, custom view components would be supported only through installed uiplugins and is the option to get total control of the action node rendering.

schemas/component/ui.schema
```diff
@@ -1,113 +1,189 @@
...
+                        },
+                        {
+                            "type": "array",
+                            "items": {
+                                "$ref": "#/definitions/SubmenuOptions"
+                            }
+                        },
+                        {
+                            "const": false
+                        }
```
tdurnford, line 170: Does `"const": false` need to be included in the `oneOf`? We can assume the submenu property is falsy if it's undefined.

a-b-r-o-w-n: I think if this is undefined, we would group the action in the Custom Actions menu item. But that's a good question. @yeze322 do you have any thoughts?

.editorconfig
```diff
@@ -0,0 +1,12 @@
...
+[*]
+insert_final_newline = true
+
+[*.json]
+indent_style = space
+indent_size = 4
+
+[*.schema]
+indent_style = space
+indent_size = 4
```
chrimc62, line 12: You didn't want to check this in did you?

schemas/ui/v1.0/ui.schema
```diff
@@ -0,0 +1,189 @@
...
+                "header": {
+                    "title": "Node Header",
+                    "description": "Text used in the node header.",
+                    "type": "string"
+                },
+                "body": {
+                    "title": "Node Body",
+                    "description": "TODO",
+                    "type": "string"
+                },
```
yeze322, line 41: Needs another property 'footer' here, for example the `SetProperty` $kind.

schemas/component/ui.schema
yeze322: When "label" is `undefined` or `''`, will the default label resolver generate a fallback label string?
If so, I think `false` is better to indicate that user whats to hide the label.

yeze322:
As Chris told me, we will serialize whole Flow schema into the ui.schema json (covered every $kinds), I think it's a bit challenging.

Because of node appearances are quite flexible, we have to consider how can we describe styles, data transformers in the schema json. As an example, below is the flow schema of `SetProperty`, it imports a React component and includes a body function.
```ts
[SDKKinds.SetProperty]: {
  widget: 'ActionCard',
  body: (data) => <PropertyAssignment property={data.property} value={data.value} />,
}
```
So in short term, it's not likely we can serialize every $kind in Flow schema into UI schema. We have to rely on some builtin TS schema defs. Chris also considered it a lower priority work compares with the 'form' and 'flyout' parts.

___
What I want to suggest here is, can we just let UI schema include a subset of $kinds in the 'flow' part, make the UI schema override the builtin Flow schema?

The benefits are:
1. Flow schema can keep simple, only 'header', 'body', 'footer', 'icon', 'color'. We only provide liminited customization ability.
2. Users can define custom node appearances, this will serve the 'Custom Action' story well.

yeze322:
What I want to suggest is can we just let UI schema include a subset of $kinds in the 'flow' part? The difference is the UI schema overrides the builtin Flow schema instead of serializes all of it.

The benefits I can imagine are:
1. Flow schema can keep simple. If we only provide limited customization ability, current schema defined by Andy is descriptive enough.
2. Serve the 'Custom Action' story. Users can define customize Action node appearances in ui.schema file, this will be useful.

@cwhitten @a-b-r-o-w-n what's your opinion on that?

chrimc62:
I suspect the .uischema for SDK components should be delivered as part of the SDK components themselves.  That way versioning of default .uischema and .schema stay in sync.  

---
In reply to: [656485907](https://github.com/microsoft/botframework-sdk/pull/5955#issuecomment-656485907) [](ancestors = 656485907)

schemas/ui/v1.0/ui.schema
```diff
@@ -0,0 +1,189 @@
...
+    "title": "UI Schema",
+    "description": "Visual customization options for BotFramework SDK components.",
+    "type": "object",
+    "additionalProperties": true,
+    "properties": {
+        "flow": {
+            "title": "Flow UI Options",
+            "description": "UI Options object describing flow UI customizations for this $kind.",
+            "$ref": "#/definitions/FlowUIOptions",
+            "additionalProperties": false
```
chrimc62, line 13: These are the wrong place and validation will complain.  You need to put them as a peer to properties which is found in your FlowUIOptions.  Same for the other sections.

.editorconfig
a-b-r-o-w-n: I did mean to check this in. My editor was trained to do 2 spaces instead of 4.

chrimc62: The problem is that this is a "public" space.  So anyone looking at schema will also find this there which I don't think is a good idea.

schemas/component/ui.schema
a-b-r-o-w-n: We do use false elsewhere to indicate hiding the label. In this case, there is no option to hide the label. We could use `null` instead. What are your thoughts?

.editorconfig
a-b-r-o-w-n: I'm not sure I follow? This configuration only affects this git repo (and only with editors that support + use editor config). It is just a formatting configuration for *.schema files that are checked in to this repository.

a-b-r-o-w-n:
We decided to punt on the flow portion of the schema for now. We'd like to use the expression engine to allow more dynamic values and work on refining the scope of what the flow configuration should be.

PR #6020, opened 2020-09-01
Initial csharp parser commit

### LINKED ISSUE #6013: Generate antlr C# runtime and port luParser.js
Create a new nuget package.
Generate C# antlr runtime from the same Lexer and Parser Files in the JS implementation
Port [luParser.js](https://github.com/microsoft/botframework-sdk/blob/main/parsers/LU/JS/packages/lu/src/parser/lufile/luParser.js)

axelsrz:
fixes #6013


parsers/LU/CS/LUParser/Program.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
```
munozemilio, line 1: Not needed

munozemilio (commented):
CIL

PR #6047, opened 2020-09-18
Adding intellisense to uiSchema

LouisEugeneMSFT:
Fixes https://github.com/microsoft/BotFramework-Composer/issues/4186

## Proposed Changes

In R10, we introduced Intellisense to Composer fields.
For this first release, the mappings between components and valid Intellisense scopes* were hardcoded.
_*Scopes refer to the type of results that Intellisense returns, such as "user-variables"._

In R11, we would like to have the uiSchema dictates this mapping.

Below is a screenshot of how overwriting the uiSchema enables custom Intellisense scopes.
 MicrosoftTeams-image

### DIFF
--- a/schemas/ui/v1.0/ui.schema
+++ b/schemas/ui/v1.0/ui.schema
@@ -57,7 +57,7 @@
                     "items": {
                         "type": "string"
                     },
-                    "examples": ["expressions","user-variables"]
+                    "examples": [["expressions","user-variables"]]
                 },
                 "label": {
                     "title": "Label",
--- a/schemas/ui/v1.0/ui.schema
+++ b/schemas/ui/v1.0/ui.schema
@@ -50,6 +50,15 @@
                         "type": "string"
                     }
                 },
+                "intellisenseScopes": {
+                    "title": "Intellisense scopes",
+                    "description": "An array of valid scopes to help Intellisense return the correct type of results.",
+                    "type": "array",
+                    "items": {
+                        "type": "string"
+                    },
+                    "examples": ["expressions","user-variables"]
+                },
                 "label": {
                     "title": "Label",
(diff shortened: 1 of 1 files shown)


schemas/ui/v1.0/ui.schema
```diff
@@ -50,6 +50,15 @@
...
                     }
                 },
+                "intellisenseScopes": {
+                    "title": "Intellisense scopes",
+                    "description": "An array of valid scopes to help Intellisense return the correct type of results.",
+                    "type": "array",
+                    "items": {
+                        "type": "string"
+                    },
+                    "examples": ["expressions","user-variables"]
```
a-b-r-o-w-n, line 60:
```suggestion
                    "examples": [["expressions","user-variables"]]
```

This needs to be an array of arrays instead.

a-b-r-o-w-n (changes requested):
@chrimc62 do you have any opinion on this?

schemas/ui/v1.0/ui.schema
LouisEugeneMSFT: good catch!

a-b-r-o-w-n (approved):

schemas/ui/v1.0/ui.schema
chrimc62: I'm confused.  That is not what you have modeled in the schema.  Why an array of arrays?  And if so the schema and description should reflect that.  The items should be of type array with items of string.

---
In reply to: [492409522](https://github.com/microsoft/botframework-sdk/pull/6047#discussion_r492409522) [](ancestors = 492409522)

chrimc62 (changes requested):
:clock1:

chrimc62 (approved):
:shipit:

Issue #6025, opened 2020-09-04
Writeup skills functional spect


gabog:
The working draft can be found here: https://github.com/microsoft/botframework-sdk/blob/clearab/skill-spec/specs/testing/skills/SkillsFunctionalTesting.md

cleemullins:
Discussed w/ @gabog and the spec is in progress, but not needed to land as part of R11 (no code deliverable). Pushing to R12.

mrivera-ms:
Per conversation with @gabog, this item spans several milestones. Not to be completed in R13.

PR #6074, opened 2020-10-08
Josh/matrix

Virtual-Josh:
Related to #6025 

Adding matrices for tracking information for channels and adaptive card support


Virtual-Josh:
I expect these matrices to change as we I continue to gain more understanding of what's needed here.

gabog (approved):
Just add (DRAFT) to the title if you can

PR #6093, opened 2020-10-28
Update default basemodel in  nlr_versions.json

tsuwandy:
Fixes #
Ning’s evaluations show that the 6 layer is the best model as default base model.

## Proposed Changes

  - 
  -
  -


hcyang:
Hi, Tien, Dave has just uploaded new .zip files. Can you work with him on this nlr_versions.json file? Also maybe we should rename nlr to basemodel?

daveta (approved):

PR #6208, opened 2021-02-19
Add Action Policy definitions to component schema

EricDahlvang:
Fixes https://github.com/microsoft/botframework-sdk/issues/6207

thanks @chrimc62 !

### DIFF
--- a/schemas/component/component.schema
+++ b/schemas/component/component.schema
@@ -9,6 +9,48 @@
         },
         {
             "properties": {
+                "$policies": {
+                    "type": "array",
+                    "items": { 
+                        "oneOf": [
+                            {
+                                "type": "object",
+                                "properties": {
+                                    "type": {
+                                        "type": "string", 
+                                        "string": "mustInclude"
+                                    },
+                                    "kinds": {
+                                        "type":"array",
+                                        "items": {
+                                            "type": "string"
+                                        }
+                                    }
+                                },
+                                "required": [
+                                    "type",
+                                    "kinds"
+                                ]
+                            }, 
+                            {
+                                "type": "object",
+                                "properties": {
+                                    "type": {
+                                        "type": "string", 
+                                        "enum": [
+                                            "lastAction",
+                                            "interactive",
+                                            "notInteractive"
+                                        ]
+                                    }
+                                },
+                                "required": [
+                                    "type"
+                                ]
+                            }
+                        ]
+                    }
+                },
                 "$role": {
                     "title": "$role",
                     "description": "Defines the roles played in the dialog schema one of [expression|interface|implements($kind)|extends($kind)].",
--- a/schemas/component/component.schema
+++ b/schemas/component/component.schema
@@ -18,7 +18,7 @@
                                 "properties": {
                                     "type": {
                                         "type": "string", 
-                                        "string": "mustInclude"
+                                        "string": "requiresAction"
                                     },
                                     "kinds": {
                                         "type":"array",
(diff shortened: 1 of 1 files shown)


schemas/component/component.schema
```diff
@@ -9,6 +9,48 @@
...
+                "$policies": {
+                    "type": "array",
+                    "items": { 
+                        "oneOf": [
+                            {
+                                "type": "object",
+                                "properties": {
+                                    "type": {
+                                        "type": "string", 
+                                        "string": "requiresAction"
```
chrimc62, line 21: >string [](start = 41, length = 6)

const

schemas/component/component.schema
```diff
@@ -9,6 +9,48 @@
...
+                                            "interactive",
+                                            "notInteractive"
+                                        ]
+                                    }
+                                },
+                                "required": [
+                                    "type"
+                                ]
+                            }
+                        ]
```
chrimc62, line 51: Why did you make this an object and not just an enum?  Do you anticipate needing to add other properties to these policies?

chrimc62:
When this is checked-in, I will update dialog:merge to both validate and expand kinds in the policy.  This will catch errors where people do typos or don't include a required package.

schemas/component/component.schema
EricDahlvang: @chrimc62 This just does not feel right to me:

```
"$policies": [
	{
		"type": "mustInclude",
		"kinds": [
			"Microsoft.xyz"
		]
	},
	"lastAction"
]
```

Consumption of a json array of objects guaranteed to have 'type', is less complex.  Do you think this is a mistake?

chrimc62 (changes requested):
:clock1:

schemas/component/component.schema
EricDahlvang: #fixed

schemas/component/component.schema
chrimc62: I'm fine either way.  The advantage of an enum is less typing and immediate discovery in the list for customers writing schemas.  The disadvantage is mainly on the consumption side where you have to deal with strings and objects.  The only consumers are likely to be us so I tend to favor making it easier for customers writing schemas even if we have to deal with both, but either solution is fine.  

---
In reply to: [579387313](https://github.com/microsoft/botframework-sdk/pull/6208#discussion_r579387313) [](ancestors = 579387313)

chrimc62 (approved):
:shipit:

schemas/component/component.schema
chrimc62: Part of the reason I don't think just having objects is easier to consume is because you have to deal with different shape objects.  Strings are just a simple version of that and have the advantage of simpler to discover and type.

---
In reply to: [579391141](https://github.com/microsoft/botframework-sdk/pull/6208#discussion_r579391141) [](ancestors = 579391141,579387313)

schemas/component/component.schema
```diff
@@ -9,6 +9,48 @@
...
+                "$policies": {
+                    "type": "array",
+                    "items": { 
+                        "oneOf": [
+                            {
+                                "type": "object",
+                                "properties": {
+                                    "type": {
+                                        "type": "string", 
+                                        "const": "requiresAction"
```
chrimc62, line 21: >requiresAction [](start = 50, length = 14)

This should either be requires or requiresKinds or kinds should become actions.

EricDahlvang: changed to requiresKind

schemas/component/component.schema
EricDahlvang: Change to object

Thanks for this feedback @chrimc62

PR #6253, opened 2021-04-06
Fixed Swagger spec

ItalyPaleAle:
In the Swagger spec, the operationId `Conversations_GetConversationMembers` was defined twice, and that caused an issue with the Swagger CLI's validation. One of the two instances should have been `Conversations_GetConversationMember` instead.


ItalyPaleAle:
I made a second commit that also added:

- `"x-nullable": true` to all date fields
- `"x-omitempty": true` to all arrays

When generating a client with go-swagger, these options are required to ensure that zero dates or empty arrays are not included in the resulting JSON.

Not only including them would be a mistake, but they actually crash the Bot Framework Emulator app.

PR #6275, opened 2021-04-24
Add Orchestrator dispatch sample (simple with skill, qna, luis)

scheyal:


tsuwandy (approved):
Don’t we want sample to be in the sample repo?

PR #6321, opened 2021-06-16
Skill Manifest 2.2

### LINKED ISSUE #6320: Update Skill manifests to support URI-References
[v 2.1 of the skill manifest schema](https://github.com/microsoft/botframework-sdk/blob/main/schemas/skills/v2.1/skill-manifest.json) only supports absolute URI format for the following fields:

- [privacyUrl](https://github.com/microsoft/botframework-sdk/blob/main/schemas/skills/v2.1/skill-manifest.json#L42)
- [iconUrl](https://github.com/microsoft/botframework-sdk/blob/main/schemas/skills/v2.1/skill-manifest.json#L55)
- [languageModel/url](https://github.com/microsoft/botframework-sdk/blob/main/schemas/skills/v2.1/skill-manifest.json#L358)

We have a scenario where we would like to be able to package a manifest and related assets in a single compressed file so it can be downloaded/uploaded for local processing. 

To support offline processing, we will be changing the format for the properties above to accept `uri-reference`

For example, the current definition for [languageModel/url](https://github.com/microsoft/botframework-sdk/blob/main/schemas/skills/v2.1/skill-manifest.json#L358) would change from:

```json
"url": {
  "type": "string",
  "format": "uri",
  "description": "A URL for the location of the language model file"
},
```

to 

```json
"url": {
  "type": "string",
  "format": "uri-reference",
  "description": "A URI for the location of the language model file"
},
```

## Manifest processing
Based on the [RF3986](https://datatracker.ietf.org/doc/html/rfc3986#section-4.1) definition for `uri-reference`, the applications processing the manifest instance should interpret the fields above as follows:
1. If the URI-reference's prefix does not match the syntax of a scheme (e.g. `file`:, `https:`, etc.) followed by its colon separator, then the URI-reference is a relative reference.
2. Absolute URIs should be resolved as specified in the field
3. Relative URIs should be resolved by combining the URI for the manifest with the relative URI specified in the field

image

## Manifest versioning implications
This change is backward compatible so manifests created against the 2.1 version of the skill manifest schema should still be valid against this updated version (the target new version will be 2.2)

## Question for reviewers
Do we want to support URI references for [endpointUrl](https://github.com/microsoft/botframework-sdk/blob/main/schemas/skills/v2.1/skill-manifest.json#L180)? I can't think of a use case where we would have the bot endpoint relative to the manifest location but just asking since we are going to create a 2.2 manifest and there's an opportunity to expand this.

### LINKED ISSUE #6323: Use "title" field in skill manifest
## Issue

The current skill manifest has lots of properties of the form 
`"name": {
            "type": "string",
            "description": "Name of the skill"
        }`

## Proposed change
To ease localization, they should be given a "title" field as well, e.g. 
`"name": {
            "type": "string",
            "title": "Skill name",
            "description": "Name of the skill"
        }`

## Component Impact 

Just the skill manifest.

## Customer Impact 

This will let Composer's localization system label fields correctly in the skill-export dialog and anywhere else these field names occur.

gabog:
Fixes #6320
Fixes #6323

In this PR: 
* Created a new v2.2 folder for the new version of the manifest
* Updated the `privacyUrl`, `iconUrl` and `languageModel/url` properties to accept `uri-reference`:
* Added new [for relative URLs ](https://github.com/microsoft/botframework-sdk/tree/gabog/SkillManifest2_2/schemas/skills/v2.2/samples/relativeUris)
* Updated other samples to reference manifest 2.2
* Fixed schema validation issue for 2.1 and bumped the version to 2.1.1
* Added test project that validates samples against the manifest schema.

Relevant changes to manifest:
image

image

image

FYI: @sbacha, @babrekel


luhan2017 (approved):
Lgtm

schemas/skills/v2.2/skill-manifest.json
```diff
@@ -0,0 +1,372 @@
...
+                "intents": {
+                    "type": "array",
+                    "uniqueItems": true,
+                    "description": "A list of the skill's top level intents that can be used by the consumer to filter out language files.",
+                    "items": {
+                        "type": "string"
+                    }
+                }
+            },
+            "additionalProperties": false
```
beyackle, line 104: Would you mind adding a "title" field to each of these alongside the descriptions? It can just be the same as the key but with proper capitalization and spacing (e.g. `Dispatch Models` for `dispatchModels`). I think these only need to be at the eleven top-level properties here.

gabog: @beyackle, I added title to all the properties, can you please review before I merge?

Thanks

schemas/skills/v2.2/skill-manifest.json
```diff
@@ -0,0 +1,406 @@
+{
+    "$id": "https://schemas.botframework.com/schemas/skills/v2.1/skill-manifest.json",
```
EricDahlvang, line 2: v2.2?

gabog: Right!, thanks!!

beyackle (approved):

PR #6569, opened 2023-01-25
Removes unnecessary assignment in parsePredictedResultToLu

mikemikhaylov:
Fixes #

## Proposed Changes

  - Removes unnecessary assignment in `parsePredictedResultToLu`, since `updatedText` is not used

## Testing

—


tracyboehrer:
@mikemikhaylov Apologies as I realize this is quite old.  We will not be merging due to deprecation of Bot Framework SDK EOM December 2025.

